PR #569, opened 2012-02-14
Hide vertical scrollbar in messages.

shirosaki:
Some messages have meaningless vertical scroll bar when showing quoted text.
Here is a example on Windows XP with Thunderbird 11.0.
http://i.imgur.com/gJzsC.png

`iframe.style.height` seems 1 px shorter. This patch is workaround to hide scroll bar.


SkyCorpFr:
Hello,
I can see 2 vertical scroll bars in some emails, which is quite annoying, and I believe this is the same bug.
Have you find a way to remove this additional vertical scroll bar?

PR #715, opened 2013-03-14
RFE: message info in the header

mook:
(This pull request is a crappy proof of concept, it's not expected to actually land)

It would be useful to include information about the currently-displayed message in the header when it's partially scrolled off (i.e. you can't read the header information while in the middle of reading the message).  For me, it's mostly useful for sender information (who is this bugmail from?).

(Running the extension with the given commit sort of shows what I mean, except with horrible UI)


Standard8:
This might be interesting to explore sometime in the future after the current rewrite. However given the age and no further development I think I will close it out for now.

PR #846, opened 2014-08-18
adding unread and read options to more dropdown in message

disassembler:
When installing conversations extension, I noticed the mark as read/unread buttons disappeared from the more actions on the individual messages. This PR adds these options back in.


Standard8:
Given the age of this, and the fact the current rewrite is obsoleting the code here. I will close this. Quite happy to discuss adding something similar to the rewritten ui if it makes sense.

PR #899, opened 2014-12-10
updated plugin for DKIM Verifier

lieser:
Simplified the plug-in.

The new plug-in version needs at least the 1.3.0 version of the DKIM Verifier.
This update is not time critical as the old plug-in version works also with the new DKIM Verifier version 1.3.0, thought there can occur some issues if the DKIM result is saved and both classic and conversation view is used.


protz:
Thanks. Sorry about the delay, life got really busy since you submitted this. This will be in the next release.

Issue #885, opened 2014-11-12
Regression: Contact tooltip only showing on "Me"


pianoslum:
I just noticed: On the 2.9+ branch the tooltip (when you hover with the mouse over a contact's name in conversation view) doesn't show up (except on "me"). I'm quite surprise as I didn't notice that before! Can someone confirm? It's also possible I srewed up something with my git ;)

It doesn't seem to be due to a recent commit as I tried one from August as well.

pianoslum:
I don't understand too much of that CSS/JQuery/whatever-mixture, but I pinpointed the error to one line:
If `overflow: hidden` is omitted in skin/conversation.css:217 then the popup is shown (the content is a little disarranged, but this could be another issue!?).

Can you make sense of this?

protz:
Seems like the right fix to me. What's wrong with the contents? (I 
couldn't see any defect, except that the github sender has an email 
address that's too long and overflows past the tooltip.)

~ jonathan

protz:
Wait no the comment is right, the text-ellipsis needs this to work. I 
need to think about it a little more.

~ jonathan

Issue #908, opened 2015-01-22
wrong display auf quotes and text inbetween


tofi86:
mail sources:

```
This is a multi-part message in MIME format.
--------------050108090809010704030100
Content-Type: text/plain; charset=utf-8; format=flowed
Content-Transfer-Encoding: 8bit

> Was haltet ihr von den folgenden VorschlÃ¤gen (alle noch frei):
>
>   * suggestion 1
>
+1 von BD
>
>   * suggestion 2
>   * suggestion 3
>   * suggestion 4
>
+/- 0 von BD


--------------050108090809010704030100
Content-Type: text/html; charset=utf-8
Content-Transfer-Encoding: 8bit

<html>
  <head>
    <meta content="text/html; charset=utf-8" http-equiv="Content-Type">
  </head>
  <body bgcolor="#FFFFFF" text="#000000">
    <br>
    <blockquote cite="mid:54BFDC89.1010701@domain.de"
      type="cite"> Was haltet ihr von den folgenden VorschlÃ¤gen (alle
      noch frei):<br>
      <ul>
        <li>suggestion 1</li>
      </ul>
    </blockquote>
    +1 von BD
    <blockquote cite="mid:54BFDC89.1010701@domain.de"
      type="cite">
      <ul>
        <li>suggestion 2<br>
        </li>
        <li>suggestion 3</li>
        <li>suggestion 4</li>
      </ul>
    </blockquote>
    +/- 0 von BD<br>
    <br>
  </body>
</html>
```

Original Thunderbird rendering (correct):
![tb_original](https://cloud.githubusercontent.com/assets/365118/5853629/92bbeb5e-a225-11e4-945b-8bf731b978ab.png)

Conversation View Plugin (incorrect):
![conversation_view](https://cloud.githubusercontent.com/assets/365118/5853633/9c982e8a-a225-11e4-861c-0465ce262b2b.png)

protz:
Did Conversation really re-order the text?!!

If that's indeed the case, please forward the message to me using Tools > Forward As > Attachment. My email address is on my github profile page.

Thanks,

Jonathan

PR #926, opened 2015-03-20
Use 'for ... of' instead of deprecated 'for each ... in'

pianoslum:
As the title says, 'for in' is deprecated, so I replaced all occurrences.
As this affects many files we should test... well everything ;-)

There's a second thing:
let blocks are deprecated. This happens twice in std-lib and once in Conversations:
contact.js:273 What does this do and how can we fix this? And where does all that hatred come from?

For #885 I am still waiting for a feedback from you. I don't say this to yield stress, but just to ensure that we are not in a "deadlock" ;-)


protz:
Haha. Amusing comment in `contact.js:273`. I'll fix this it's easy.

`let`-blocks were nice, but never took off in the TC36 committee, so I don't think Mozilla can afford to keep these Mozilla-only JS constructs. It just makes no sense.

protz:
Thanks! There's some subtleties related to the `Iterator` thing that I've been using and which is also non-standard. I'm reading https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols to figure out what to do in those cases. It's related to that `Iterator.symbol` thing you mentioned earlier.

protz:
Quote from MDN on the page of `for each ... in`: « for...of is a similar statement that iterates over the property values but there is some behavior that is different from for each...in statements. »

protz:
Ok, here's what I think we need to do. There are multiple use-cases:

* `for each (let [, x] in Iterator(foo))`: if `foo` is an [Iterable Object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols), replace with `for (let x of foo)` (much simpler). Works for: arrays, DOM collections.
* `for each (let [, x] in fixIterator(foo))`: rewrite into `for (let x of fixIterator(foo))` (see documentation for http://mxr.mozilla.org/comm-central/source/mailnews/base/util/iteratorUtils.jsm#39)
* all other cases: study each case carefully and try to find alternate ways (for instance, `for each (let [x, y] in Iterator(foo))` cannot be replaced by `Object.keys(foo).forEach(...)` blindly, because the behavior is not the same for arrays...

protz:
More rewriting rules. I can't seem to find a way to easily iterate on key-value pairs in an array...

* `for each (let [i, y] in Iterator(array))`: `array.forEach(function (y, i) {...`
* some other cases are better served by using ES6 `Map`s (see my inline comment, which probably also applies to other functions in that file)

I've skimmed through the diff and I believe that covers most of the use-cases. I'm sorry you did all the rewriting work already and some things need to be updated, but I believe that this should be pretty easy and relatively mechanical.

If you're willing to do this, I'd be happy to scrutinize your pull request to make sure everything looks right. Furthermore, if you have doubts in some cases, just flag them and I'll happily rewrite them myself :)

pianoslum:
Ok, I will look into this and ask you if there are any problems!

protz:
Awesome! I'm on to the attachments / offline issue.

content/stub.xhtml
```diff
@@ -392,7 +392,7 @@
     let oldPrint = window.print;
 
     function printConversation(event) {
-      for each (let { message: m } in Conversations.currentConversation.messages) {
+      for (let { message: m } of Conversations.currentConversation.messages) {
```
pianoslum: @protz What is this `{ message: m }` Do I have to take care about that?

modules/prefs.js
```diff
@@ -67,7 +67,7 @@ function PrefManager() {
   this.compose_in_tab = prefsService.getBoolPref("compose_in_tab");
   // This is a hashmap
   this.monospaced_senders = {};
-  for each (s in this.split(prefsService.getCharPref("monospaced_senders")))
+  for (s of this.split(prefsService.getCharPref("monospaced_senders")))
```
pianoslum: @protz Why is there no `let`? Normally, I use `for (let variable of ...`

content/stub.xhtml
protz: It's an ES6 destructuring assignment, which binds variable `m` which is used...

content/stub.xhtml
```diff
@@ -392,7 +392,7 @@
...
         m.dumpPlainTextForPrinting();
```
protz: right here.

modules/prefs.js
protz: That's a mistake on my side, and I'm surprised that the strict mode didn't catch that.

protz: Good catch :)

pianoslum:
I did some further simplifications, but I'm still having problems:
Why is it sometimes ok to change `for (let [,x] of Iterator(foo)` to `for (let x of Iterator(foo)` and sometimes not?
E.g. in stub.xhtml:66 this leads to an error.

Right now I do most of the things by trial&error, but is there a better way? Another problem is that I can't see if a variable is an array, an object, an iterable or whatever.
Any ideas?

protz:
On line 66 you should change to `for (let menu of menus)`, because `menus` is a DOMCollection which implements the `Iteratable` interface, meaning that you don't need to cast it into an `Iterator`...

protz:
I don't think we should `Iterator` anywhere anymore, it's going away too. From the rewriting rules I gave above, all uses boil down to either a native `for-of` loop without `Iterator`, or a `for-of` loop with `fixIterator` (because of XPCOM madness), or something different (a `Map`, an `array.forEach`, etc).

Do you want me to review something at this stage?

protz:
As a head-up warning, I just had to fix a few for each's in the 
codebase, while fixing #908. You may have conflicts, sorry about that.

Also, I think I uncovered a case of miscompilation of a for/each loop 
(see https://bugzilla.mozilla.org/show_bug.cgi?id=1148944), so it's 
definitely a great thing to get rid of these, the more we wait, the more 
random breakage we're going to get.

~ jonathan

content/stub.completion-ui.js
```diff
@@ -84,15 +84,16 @@ ContactIdentityCompleter.prototype = {
     }
     // and since we can now map from contacts down to identities, map contacts
     //  to the first identity for them that we find...
-    matches = [val.NOUN_ID == Gloda.NOUN_IDENTITY ? val : val.identities[0]
-               for each ([iVal, val] in Iterator(contactToThing))];
+    matches = Array.prototype.map.call(contactToThing, function(val) {
+      return val.NOUN_ID == Gloda.NOUN_IDENTITY ? val : val.identities[0];
+    });
```
protz: While all the above fixes seem right, this one seems weird.

Here's a suggestion. For all the cases where Iterator is called with an Object that's abused as a key/value store, use entries.

* Copy/paste somewhere the entries function from https://esdiscuss.org/topic/es6-iteration-over-object-values
* Rewrite `for each ([k, v] in Iterator(obj))` into `for (let [k, v] of entries(obj))` and everything should be alright and easy!

content/stub.completion-ui.js
```diff
@@ -84,15 +84,16 @@ ContactIdentityCompleter.prototype = {
...
 
     let rows = [asToken(
                   match.pictureURL(),
                   match.contact.name != match.value ? match.contact.name : null,
                   match.value,
                   match.value
-                ) for each ([iMatch, match] in Iterator(matches))];
+                ) for ([iMatch, match] of matches)];
```
protz: Same remark here I think.

content/stub.completion-ui.js
```diff
@@ -133,13 +134,16 @@ ContactIdentityCompleter.prototype = {
       this.identityCollection =
         this.contactCollection.subCollections[Gloda.NOUN_IDENTITY];
 
-      let contactNames = [(c.name.replace(" ", "").toLowerCase() || "x") for each
-                          ([, c] in Iterator(this.contactCollection.items))];
+      let contactNames = this.contactCollection.items.map(function(c) {
+        return c.name.replace(" ", "").toLowerCase() || "x";
+      });
```
protz: Here, I just looked up the source code, items is an array, so that's good.

content/stub.completion-ui.js
```diff
@@ -133,13 +134,16 @@ ContactIdentityCompleter.prototype = {
...
       // if we had no contacts, we will have no identity collection!
       let identityMails;
-      if (this.identityCollection)
-        identityMails = [i.value.toLowerCase() for each
-                         ([, i] in Iterator(this.identityCollection.items))];
+      if (this.identityCollection) {
+        identityMails = Array.prototype.map.call(this.identityCollection.items, function(i) {
+          return i.value.toLowerCase();
+        });
+      }
```
protz: why not map here? I suspect identityCollection is a glodaCollection too, meaning its items property is an array, meaning you can call map on it

content/stub.completion-ui.js
```diff
@@ -190,12 +194,14 @@ ContactIdentityCompleter.prototype = {
...
-                    dude.value
-                  ) for each ([iDude, dude] in Iterator(possibleDudes))];
+      let rows = Array.prototype.map.call(possibleDudes, function(dude) {
+          return asToken(
+            dude.pictureURL(),
+            dude.contact.name != dude.value ? dude.contact.name : null,
+            dude.value,
+            dude.value
+          )
+      });
```
protz: same here, `possibleDudes.map` should work

content/stub.completion-ui.js
protz: Amendment: since `matches` is an array, then you want `for match of matches`. The elements in your array are not pairs, I think...

content/stub.completion-ui.js
protz: Meaning: `matches = [... for ([iVal, val] of entries(contactToThing))]`. I should really submit a patch to have this exposed so that we don't maintain a copy of it in the Conversations tree.

content/stub.completion-ui.js
```diff
@@ -299,27 +305,27 @@ function setupAutocomplete(to, cc, bcc) {
...
-      let li = document.createElement("li");
-      li.setAttribute("title", email);
-      li.textContent = name;
-      let span = document.createElement("span");
-      span.classList.add("recipientListSeparator");
-      span.textContent = sep;
-      list.insertBefore(li, marker);
-      list.insertBefore(span, marker);
-    }
+    aData.forEach(function(v, i) {
```
protz: `aData.forEach(function ({ name, email}, i) { ...`

content/stub.completion-ui.js
```diff
@@ -230,7 +236,7 @@ function peopleAutocomplete(query, callback) {
   let add = function(person) {
     let photos = person.getProperty("photos");
     let thumb;
-    for each (let photo in photos) {
+    for (let photo of photos) {
```
protz: I think this is dead code now...

content/stub.completion-ui.js
```diff
@@ -238,7 +244,7 @@ function peopleAutocomplete(query, callback) {
     }
 
     let suggestions = person.getProperty("emails");
-    for each (let suggestion in suggestions)
+    for (let suggestion of suggestions)
```
protz: Same here.

content/stub.compose-ui.js
```diff
@@ -684,8 +684,9 @@ ComposeSession.prototype = {
       draft: function ({ to, cc, bcc }) {
         let makeTokens = function (aList) {
           let [list, listEmailAddresses] = parse(aList);
-          return [asToken(null, item, listEmailAddresses[i], null)
-            for each ([i, item] in Iterator(list))];
+	  return Array.prototype.map.call(list, function(item, i) {
```
protz: `list.map`?

protz: You seem to hesitate between rewriting `[f(x, i) for each ([i, x] in Iterator(someArray))]` into either:

- `someArray.map(function (x, i) f(x, i))` or
- `[f(x, i) for ([i, x] of someArray)]`.

Since https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#The_iterable_protocol tells us that `Array` implements the iteration protocol, the two above should be equivalent. I'd rather go for the former, what do you think?

content/stub.xhtml
```diff
@@ -62,9 +62,9 @@
     })(jQuery);
 
     function clearMenu() {
-      let menus = document.getElementsByClassName("menu");
-      for each (let [, menu] in Iterator(menus))
+      Array.prototype.forEach.call(document.getElementsByClassName("menu"), function(menu) {
```
protz: for (let menu of menus)

content/stub.xhtml
```diff
@@ -595,7 +590,7 @@
         let items = data.splice(0, n);
         let nodes = $("#popularContactTemplate").tmpl(items);
 
-        for each (let [i, data2] in Iterator(items)) {
+	items.forEach(function(data2, i) {
```
protz: indent

content/stub.compose-ui.js
protz: My bad, for-of doesn't give you the index when iterating on an array. So yeah there are some cases where you do have to use map.

modules/assistant.js
```diff
@@ -263,7 +263,7 @@ let Customizations = {
 
       let vFolder = VirtualFolderHelper.wrapVirtualFolder(smartInbox);
       let searchFolders = {};
-      for each (let [, folder] in Iterator(vFolder.searchFolders)) {
+      for (let folder of Iterator(vFolder.searchFolders)) {
```
protz: this is an array, you can drop the `Iterator`

modules/contact.js
```diff
@@ -84,7 +84,7 @@ ContactManager.prototype = {
     // Might change in the future... who knows? ...
     let key = email;
     let cache = function _cache (name, contact) {
-      for each (let [, email] in Iterator(contact.emails)) {
+      for (let email of Iterator(contact.emails)) {
```
protz: same here

modules/conversation.js
```diff
@@ -173,7 +173,7 @@ let OracleMixIn = {
         //  message if all messages are read), and we expand all unread messages
         //  + the last one (which will probably be unread as well).
         if (this.scrollMode == Prefs.kScrollUnreadOrLast) {
-          for each (let [i, { message }] in Iterator(this.messages)) {
+          for (let [i, { message }] of Iterator(this.messages)) {
```
protz: same

modules/conversation.js
```diff
@@ -182,7 +182,7 @@ let OracleMixIn = {
         // In this mode, we scroll to the selected message, and we only expand
         //  the selected message.
         } else if (this.scrollMode == Prefs.kScrollSelected) {
-          for each (let [i, { message }] in Iterator(this.messages)) {
+          for (let [i, { message }] of Iterator(this.messages)) {
```
protz: same

modules/conversation.js
```diff
@@ -194,11 +194,11 @@ let OracleMixIn = {
 
         break;
       case Prefs.kExpandAll:
-        for each (let [, { message }] in Iterator(this.messages))
```
protz: same and same below

modules/conversation.js
```diff
@@ -289,7 +289,7 @@ function ViewWrapper(aConversation) {
   //  compare them by messageKey (not reliable), but URLs should be enough.
   this.byUri = {};
   [this.byUri[msgHdrGetUri(x)] = true
-    for each ([, x] in Iterator(this.mainWindow.gFolderDisplay.selectedMessages))];
+    for ([, x] of Iterator(this.mainWindow.gFolderDisplay.selectedMessages))];
```
protz: for x of this.main...

modules/conversation.js
```diff
@@ -383,7 +383,7 @@ Conversation.prototype = {
   // XXX this logic is weird. Shouldn't we just compare a list of URLs?
   _selectionChanged: function _Conversation_selectionChanged () {
     let gFolderDisplay = topMail3Pane(this).gFolderDisplay;
-    let messageIds = [x.messageId for each ([, x] in Iterator(this._initialSet))];
+    let messageIds = [x.messageId for ([, x] of Iterator(this._initialSet))];
```
protz: ok I suspect you somehow stopped halfway in your rewrite, so I'll just leave it here, and check a little while later... we might want to check this in after the branching for 2.9, since this is such a huge change, I'm afraid some bugs will fall under the radar before the release... what do you think?

pianoslum:
Yes, sorry that you reviewed conversations.js I should have said it clearer:
In the first iteration I substituted `for each in` by `for of`. After that I looked again into every occurrence and deleted the `Iterator` or changed the whole line...

But thanks for your annotations - I will try to implement them.

Check after 2.9 is ok by me!

Happy Easter!

protz:
No worries, let me know when you want me to do another review again! 
This just took me 10 minutes, I thought I'd take a look :)

content/stub.completion-ui.js
pianoslum: Why? Because we use the other lib now?

content/stub.compose-ui.js
pianoslum: list.map doesn't seem to work...

protz: Maybe list is sometimes null?

pianoslum:
If I substitute a for loop with a closure (array.forEach(function)) I can't use `this` inside.
Is it the right thing to say `let self = this;` somewhere and use `self` instead?

protz:
Use an ES6 closure with the syntax:

() => {
}

which captures this in a lexical manner.

pianoslum:
OK, I already started when I saw that `let self = this;` has already been used in the same file. But I can change it if you like!

If you have time you could check all the files I committed since your last review...

protz:
let self = this also works, and is perfectly fine! It's just that the 
ES6 arrow syntax didn't exist back when I started Conversations. Either 
way is fine.

I don't expect to have much time until this weekend, but I'll try to do 
a review.

I'm tempted to land this before the next release, at least we'll get 
some testing, otherwise, if this remains on -master, I'm afraid we won't 
get a lot of coverage for your changes, and it'd be a shame to see your 
good patch not being used for another 42 weeks. Thoughts?

other/oembed-addon/bootstrap.js
```diff
@@ -63,10 +63,10 @@ let hook = {
     let links = iframeDoc.getElementsByTagName("a");
     // Don't detect links in quotations.
     [x.skip = true
-      for each ([, x] in Iterator(iframeDoc.querySelectorAll("blockquote a")))];
+      for ([, x] of Iterator(iframeDoc.querySelectorAll("blockquote a")))];
```
protz: Hey, I'm still seeing Iterator here and below, am I looking at the right version?

pianoslum:
I only worked on 
content/stub.completion-ui.js
content/stub.xhtml
modules/assistant.js
modules/monkeypatch.js
modules/contact.js
modules/conversation.js

in the meantime. I thought it's better review them by and by and not the whole commits at the end. But it's up to you ;-)

I thought we have six weeks starting from March 31st - so shouldn't there be enough time? When do you want to release?
The problem is that I underestimated this task... As it's not essential, would it be better to stop here and work on "relevant" things?

protz:
Ok! I'll review these files. I just wanted to make sure I'm reviewing 
the right thing.

You're correct that it's a pretty massive (and unrewarding task). Is 
there anything else you'd rather work on?

~ jonathan

pianoslum:
I'm OK with that - it's rather if you said that there was more important work to do, ...

So is it right that we have time until 12th of May for version 2.9?

protz:
I don't see anything urgent right now. I seem to have fixed the most
important bits (attachments, messages that didn't display). Do you see
anything that needs attention before the next release?

I would need to cut 2.9 by the end of the month, because reviewing on
addons.mozilla.org takes a while and the new version needs to be approved
before the new Thunderbird goes live (otherwise it's broken for everyone).

pianoslum:
I noticed that in keycustomizations.js and message.js a JSON-format is used for saving the key configurations. I haven't found an easy way yet to iterate over JSON-date via `for of`. Any ideas?

protz:
JSON-decoded data is just an object, so you should be able to use your 
entries function, perhaps?

~ jonathan

pianoslum:
Thanks for your help!
I actually managed to get rid of ALL for..in's and Iterator's in Conversations now. Could you review the files? I don't know whether it was a mistake to edit the modules/plugins/files.js because I couldn't test them!
Should I merge or would you like to do it (when you finished checking, of course)?

Next I will go through std-lib and do the same...

protz:
Supercool. I'll try to review this shortly, and reply more thoroughly to 
your other bug reports.

protz:
And thanks again! This is a major piece of work, congrats.

content/stub.xhtml
```diff
@@ -251,15 +251,10 @@
...
-        if (k == msgNode) {
-          index = i;
-          break;
-        }
+      let index = Array.prototype.indexOf.call(msgNodes, msgNode);
+      if (index != -1) {
+        msgNode.setAttribute("tabindex", index + 2);
       }
-      if (index)
-        msgNode.setAttribute("tabindex", index+2);
```
protz: Funny. Because of the way the previous test was phrased, you're actually changing the behavior in case i == 0. But, amusingly enough, this is functionnaly equivalent, because if the first node was set to be focused, then whether it has tabindex 1 or 2 doesn't matter... I think. Anyhow, that seems more than legit :)

modules/conversation.js
```diff
@@ -383,7 +386,9 @@ Conversation.prototype = {
   // XXX this logic is weird. Shouldn't we just compare a list of URLs?
   _selectionChanged: function _Conversation_selectionChanged () {
     let gFolderDisplay = topMail3Pane(this).gFolderDisplay;
-    let messageIds = [x.messageId for each ([, x] in Iterator(this._initialSet))];
+    let messageIds = this._initialSet.map(function(x) {
+      return x.messageId;
+    });
```
protz: In Gecko, we have the shorter form `x => x.messageId` now.

modules/conversation.js
```diff
@@ -1152,12 +1159,12 @@ Conversation.prototype = {
       self._htmlPane.scrollNodeIntoView(focusedNode);
       self.messages[focusThis].message.onSelected();
 
-      for each (let [i, node] in Iterator(messageNodes)) {
+      Array.prototype.forEach.call(messageNodes, function(node, i) {
         // XXX This is bug 611957
```
protz: This comment needs to go away, then :)

modules/conversation.js
```diff
@@ -1233,7 +1240,7 @@ Conversation.prototype = {
       }
     }
     let messagesHtml = new Array(this.messages.length);
-    for each (let [i, { message: message }] in Iterator(this.messages)) {
+    for (let [i, { message: message }] of this.messages) {
```
protz: This one looks incorrect.

protz: Maybe you need `entries`?

modules/message.js
```diff
@@ -1974,7 +1976,9 @@ let PostStreamingFixesMixIn = {
       let _log = function () {
         let t1 = Date.now();
         let delta = t1 - t;
-        let args = [x for each ([, x] in Iterator(arguments))];
```
protz: Eeeeew. I don't know what this code is doing, but I'd rather get rid of `_log` entirely. Who wrote that? ;-)

modules/misc.js
```diff
@@ -107,42 +106,46 @@ function joinWordList (aElements, aInsertHtml) {
...
+    return mapping[idx][1] + ".svg";
   }
-  for each (let [k, v] in Iterator(fallbackMapping)) {
-    if (aMimeType.indexOf(k) === 0)
-      return v+".svg";
+  idx = fallbackMapping.findIndex(function([k, ]) {
+    return aMimeType.startsWith(k);
+  });
+  if (idx != -1) {
+    return fallbackMapping[idx][1] + ".svg";
```
protz: cool rewrite

modules/prefs.js
protz: This one still needs fixing.

other/oembed-addon/bootstrap.js
protz: Ok so you still needs to fix bootstrap.js

protz:
That's a heroic patch, thank you so much!

I just reviewed everything. The first two comments are for info only, nothing to change. There's bootstrap.js to fix, _log to remove (optional), one comment to remove and one missing call to entries. Also, you need to solve the merge conflicts with -master.

If you have time to fix the points above, just send me a final pull request, and I'll approve it directly and issue a new "nightly" so that we can get some testing.

If you don't have time, I can try to finish this if I find the time this week.

Thanks again, this is awesome.

pianoslum:
So I merged it - after a quick test everything seems to be fine - but I think we'll find something later ;-)

Should something related to for-loops become deprecated in the future, I'm in favour of rewriting everything to good old C-style loops ;-)

protz:
Wow, yeah, that's been a long discussion. Congratulations for the work!

PR #943, opened 2015-04-08
Add new option for "enable quick reply"

zzzeek:
While I appreciate greatly the effort and complexity that I can see went into the "Quick Reply" feature at the bottom of the conversation window,  I'd like to support the case for making this feature optional.  In my case, I prefer having a consistent set of controls for writing replies, in particular that I typically need to write my responses inline, I deal with a lot of code samples and complex reply chains,  and in general I need a lot of window room and very consistent controls.  By disabling "Quick Reply" I get the same "Reply" buttons at the bottom of the last message as that of other messages in the thread without the extra boxes and controls getting in the way.

### DIFF
--- a/content/options.xul
+++ b/content/options.xul
@@ -38,15 +38,15 @@
    - ***** END LICENSE BLOCK ***** -->
 
 <!DOCTYPE window SYSTEM "chrome://conversations/locale/options.dtd">
- 
+
 <vbox xmlns="http://www.mozilla.org/keymaster/gatekeeper/there.is.only.xul">
 
-  <setting pref="conversations.expand_who" type="radio" title="&option.expand_who;">  
-    <radiogroup>  
+  <setting pref="conversations.expand_who" type="radio" title="&option.expand_who;">
+    <radiogroup>
       <radio value="1" label="&option.expand_none;" />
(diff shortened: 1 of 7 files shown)


protz:
I think you can do that without modifying the code... (see
https://github.com/protz/GMail-Conversation-View/wiki/Tips-and-tricks:-Disabling-the-%22Quick-Reply%22-feature).
Did that work for you?

zzzeek:
I suppose I could make that recipe work, though it seems that A. clicking a checkbox is simpler and B. changing the CSS on the outside might break across releases / is inconvenient to remember for new installs?

protz:
Well,
- this is an option that I'll never test in my daily usage, meaning that
it's likely to break and cause extra maintenance trouble for me (I was
basically unable to maintain this addon for five of the past six months, so
I'm weary of adding extra burden on my back)
- I still have the hope of adding a "quick reply" feature to every message
in the conversation one day, meaning that this option would stop to make
sense.

zzzeek:
i guess there's no real way to unit test these kinds of things...javascript...

zzzeek:
okey doke then...the stylish thing works fine :)

protz:
Thanks! :)

Issue #914, opened 2015-02-11
New windows do not contain application menu toolbar items in OSX


fingres:
OSX 10.10.1
TB 31.4.0
Conversations 2.8.1

If I open a message in a new window the application menu bar only has the main TB application menu available. This results in being unable to close the window using the usual cmd+w keyboard shortcut.

![screenshot 2015-02-11 12 46 02](https://cloud.githubusercontent.com/assets/2339659/6147228/10b8f9b2-b1ed-11e4-9ed9-c2447c772bc5.png)

protz:
Correct. This is hard to fix, however, because the new window opens a bare XUL chrome that doesn't even have a `<menu>` XUL element, so even if I put "menubar" in the options for my call to `window.open`, it wouldn't know which menu to display.

The (expensive, unaffordable) way to fix this would be to write a custom XUL chrome for those pop-up conversation windows. The (right) way to fix this would be to open messenger.xul, but right now, messenger.xul cannot be used for that purpose and always displays a mail:3pane. There's plans to fix that, though, but it most likely won' thappen in the near future.

Thanks for reporting, though!

Jonathan

danduq:
OK, thanks for the explanation and for getting back to me Johnathan!

zzzeek:
how about an option so that we get "traditional" view for a message that we open in new window?

Standard8:
I believe at the moment, there's no way to open a message in a new window that has the conversations display. Hence closing for now. If it gets added back in the future, this will likely be found anyway.

PR #944, opened 2015-04-10
Allow option to open normal window w/ double click / enter

zzzeek:
I'm guessing you aren't interested in this one as it is only a temporary workaround for #914 that I'm using locally.  The window that gets opened is really unusable, as not only do keyboard controls not work but you also don't get any contextual menu on hyperlinks or anything else.    This is at least a patch folks can try out if they have the same problem.

### DIFF
--- a/modules/monkeypatch.js
+++ b/modules/monkeypatch.js
@@ -399,29 +399,31 @@ MonkeyPatch.prototype = {
       return kStubUrl + queryString;
     };
 
-    // Below is the code that intercepts the double-click-on-a-message event,
-    //  and reroutes the control flow to our conversation reader.
-    let oldThreadPaneDoubleClick = window.ThreadPaneDoubleClick;
-    window.ThreadPaneDoubleClick = function () {
-      if (!Prefs.enabled)
-        return oldThreadPaneDoubleClick();
+    if (Prefs.message_open_in_conversation) {
+      // Below is the code that intercepts the double-click-on-a-message event,
+      //  and reroutes the control flow to our conversation reader.
+      let oldThreadPaneDoubleClick = window.ThreadPaneDoubleClick;
+      window.ThreadPaneDoubleClick = function () {
+        if (!Prefs.enabled)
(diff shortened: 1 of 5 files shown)


modules/monkeypatch.js
```diff
@@ -392,29 +392,31 @@ MonkeyPatch.prototype = {
...
-      oldThreadPaneDoubleClick();
-      window.MsgOpenSelectedMessages = oldMsgOpenSelectedMessages;
-    };
-    this.pushUndo(function() window.ThreadPaneDoubleClick = oldThreadPaneDoubleClick);
+    if (Prefs.message_open_in_conversation) {
+      // Below is the code that intercepts the double-click-on-a-message event,
+      //  and reroutes the control flow to our conversation reader.
+      let oldThreadPaneDoubleClick = window.ThreadPaneDoubleClick;
+      window.ThreadPaneDoubleClick = function () {
+        if (!Prefs.enabled)
```
msprotz: You should just do `if (!Prefs.enabled || !refs.message_open_in_conversation)`, that would make your patch smaller

modules/prefs.js
```diff
@@ -65,6 +65,7 @@ function PrefManager() {
   this.hide_quote_length = prefsService.getIntPref("hide_quote_length");
   this.hide_sigs = prefsService.getBoolPref("hide_sigs");
   this.compose_in_tab = prefsService.getBoolPref("compose_in_tab");
+  this.message_open_in_conversation = prefsService.getBoolPref("message_open_in_conversation");
```
msprotz: also I think you're missing something in this file to watch for preferences changes

zzzeek: I omitted the watch for changes part because this is a monkeypatch, I'd have to write out the code to undo the monkeypatch if we're doing in-process change (right now the doc indicates restart required).     If you're interested in the patch I can take a stab at that.

Standard8:
Closing this out due to age and a feature I don't think we really need at the moment unless we get more requests for it.

PR #952, opened 2015-04-23
Reduce debug messages

### LINKED ISSUE #948: Many debug messages
I noticed some recurring debug messages. This is pretty minor, but if we have time left we could try to get rid of them:

* No chrome package registered for chrome://calendar/skin/cal-icon32.png (though there is no reference to "Conversations", this only seems to occur when the AddOn is installed. This message also pops up very often)

* Timestamp: 14.04.2015 01:04:53
Warning: Expected color but found '-moz-oddtreerowtext'.  Error in parsing value for 'color'.  Declaration dropped.
Source File: chrome://conversations/skin/tokeninput.css
Line: 38, Column: 11
Source Code:
    color: -moz-oddtreerowtext;

* Timestamp: 14.04.2015 01:04:53
Warning: Error in parsing value for 'font-style'.  Declaration dropped.
Source File: chrome://conversations/skin/tokeninput.css
Line: 113, Column: 16
Source Code:
    font-style: none;

* Timestamp: 14.04.2015 01:04:53
Warning: Unknown property '-moz-box-shadow'.  Declaration dropped.
Source File: chrome://conversations/skin/print.css
Line: 51, Column: 17
Source Code:
  -moz-box-shadow: 0 0 0 transparent;

* No chrome package registered for chrome://conversations/content/icon.png

* When switching between messages to quickly (being offline), I sometimes get something like "You failed to call signal enough times. Bad developer! Go fix your code!". While this pleases my masochistic side, a little more clarity what is actually wrong would be nice ;-) This occurs very infrequently, however.

pianoslum:
This should fix #948 and some other minor things.
I deleted the one declaration introduced by @xi (if there's an alternative, I'd be happy to add it ;-) ).

I also wrapped the Lightning-template in `if`s so that the warning (because of the missing icon) is not shown. At the moment I don't have Lightning installed, so I couldn't test whether it still works... Could you test this? Otherwise I could install it too.

I tried to update pdfjs - but the latest version has a dependency for ShellJS now. It said I should install it (`$ npm install [-g] shelljs`) but that would be rather cumbersome as every new (Conversations-)developer would have to do the same... Could you image a solution via `submodule`?

One more:
What about that `I love attachments`-option? Is it deprecated now? I could remove all traces!


bootstrap.js
```diff
@@ -230,8 +230,6 @@ function startup(aData, aReason) {
 function shutdown(aData, aReason) {
   // No need to do extra work here
   Log.debug("shutdown, aReason=", aReason);
-  // Load imports for the BOOTSTRAP_REASONS constants; should have happened before in most cases
-  loadImports();
```
protz: Why remove this? Was it causing errors?

It may be required in the scenario where Thunderbird starts with the addon disabled, then the addon is uninstalled?

protz:
Hey! Thanks again for this patch. I'll merge once you reply to my ping above.

Once this is merged, I'll most likely submit a version of Conversations for release on addons.mozilla.org. The release of Thunderbird is imminent.

- shelljs: I don't think it's excessive to require a new dependency for developers, they're developers, after all, and they only need to build pdf.js if they want to use the pdf viewer locally
- I love attachments: I think it's still relevant in some corner cases, where _allAttachments is not the same as _allUserAttachments
- I'll try to install Lightning tonight to try out your changes but apart from that everything looks fine.

You introduced tabs in your last pull request about forEach's. Please don't do that! :-)

Thanks,

Jonathan

bootstrap.js
pianoslum: The thing is the BOOTSTRAP_REASONS are defined in `Cu.import("resource://conversations/modules/config.js", global);` which isn't loaded in loadImports(). Thatswhy I thought it's useless anyway. Now it seems as it was broken before - I'll try to test that case and import the right thing if necessary.

pianoslum: I tried install-disable-restart-uninstall and there was no error...

protz: Super. Do you want me to merge now or do you want to push a commit that
updates pdf.js?

pianoslum: I don't think a new pdf version is too important...
What about the keyconfig-bug? If you knew the right Listener it could be fixed quite quickly...
Otherwise: MERGE! ;-)

protz: I'll look into it tonight! :)

protz:
Lightning is ok.

protz:
Thanks much!

PR #981, opened 2015-07-02
Fix attachments showing with PGP/MIME messages

### LINKED ISSUE #980: PGP/MIME attachments do not show in Conversations
I can perfectly reproduce this: when I send a message to e.g. myself that has encrypted PGP/MIME attachments (via Enigmail), then those attachments don't show up in Conversations. In the classic view, they do show up without a problem.

shirosaki:
I see the same issue with #980.

`this` causes an error. Instead we use `self`.
Fixes #980


protz:
Thanks @shirosaki !

Issue #967, opened 2015-06-08
Request (and offer) update UI to a more modern look


elioqoshi:
Hi there,
I actually love Conversations and use it for most quick emails I need to write.
As a designer however, it has a bit of an outdated UI, which I'd love to contribute to, its just that I'm not a coder :)
Looking forward to your thoughts!

protz:
Sure! It's actually "fairly" simple to contribute to the UI.

* Setup a development environment (see https://developer.mozilla.org/en-US/Add-ons/Setting_up_extension_development_environment)
* Then, hack on the files called content/stub.xhtml (mostly handlebars templates) and skin/conversation.css
* You'll be better off if you setup remote debugging to use the Firefox dev tools https://developer.mozilla.org/en-US/docs/Tools/Remote_Debugging/Thunderbird (otherwise, it's most likely to be a pain).

I'm definitely open to suggestions / pull requests! But if you're committed about fixing this, I think it's better to iterate on mockups first, before moving on to the actual hacking on the code?

Cheers,

Jonathan

blerner:
Potentially related to this: when updating to Tb38 on Windows today, I noticed that everything has gotten very, very white.  Apparently, the `background-color: window` decoration on the `body` of the multimessage window, and on the `#conversationHeader`, now means "white".  I changed them locally to be `transparent`, and got back the Tb37 look...

blerner:
Er, `transparent` doesn't work for the header, because scrolling content underneath it will show through.  But setting the background color to `rgb(248,248,248)` (which seems to be the computed result) does work.

boscharun:
The earlier blue background/gray header was really much better. Would it be possible to add something like a theme option from the plugin options?

P.S. Thanks a lot for this amazing plugin.

blerner:
In the meantime, I "fixed" it with a Stylish extension:
```css
@media all and (-moz-windows-default-theme) {
  browser#multimessage {
    background-color: rgb(248, 248, 248) !important;
  }

  @-moz-document url(chrome://conversations/content/stub.xhtml) {
    body { background-color: rgb(248, 248, 248) !important; }

    div#conversationHeaderWrapper > div#conversationHeader {
      background-color: #dddde0 !important;
    }
  }
}
```

boscharun:
Hi Blerner, would really appreciate if you could guide on where and how to put this code in. I am pretty sure a lot of people would want to use this.

Thanks in advance.

protz:
[...]

PR #983, opened 2015-07-09
Make the quick reply buttons conform to the system theme.

### LINKED ISSUE #982: Quick reply buttons don't conform to GTK theme
As seen in my screenshot, the quick reply buttons don't look right when using a dark theme. Also the "In new window" links are almost impossible to read.
screenshot

ctlajoie:
I am not able to test this on Windows or Mac, but I believe it will look good there too.

fixes #982

### DIFF
--- a/skin/quickreply.css
+++ b/skin/quickreply.css
@@ -64,9 +64,13 @@
 
 .quickReply .textWrap {
   margin: 10px 5px;
-  background-color: white;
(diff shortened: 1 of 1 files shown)


protz:
That's great, thanks much! If you're a CSS expert, #967 could use some love, so feel free to stop by this issue :).

Cheers,

Jonathan

PR #986, opened 2015-07-14
Change #conversationHeader color to set it off from other content.

ctlajoie:
as discussed in #967

### DIFF
--- a/skin/conversation.css
+++ b/skin/conversation.css
@@ -81,10 +81,10 @@ button {
 #conversationHeader {
   width: 100%;
   padding: 5px 10px;
-  background-color: window;
+  background-color: menu;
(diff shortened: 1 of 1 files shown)


protz:
Very cool. I don't have my linux machine around with me -- would you 
have a screenshot on Linux to make sure we don't break a platform by 
fixing another one?

ctlajoie:
I certainly made sure not to break my primary environment :)

#### Default GTK (Adwaita)
image

#### Dark theme (FlatStudioDark)
image

protz:
Very cool, thanks :)

protz:
https://addons.mozilla.org/en-US/thunderbird/addon/gmail-conversation-view/versions/ the new version has the fix, and you're mentioned in the release notes. Thanks again, and keep sending the good bugfixes :)

PR #997, opened 2015-08-19
Proper fix for issue #982

Redart98:
Changed color of backgrounds for Reply fields from buttonface to window
since it is in fact minimized window


protz:
Thanks!

Issue #1015, opened 2015-10-08
Quick-Reply is not encrypted without any warning


RalfJung:
With Conversations having some kind of support for Enigmail for decryption and signature verification, I kind of assumed that it would also have support for encryption. That, however, does not seem to be the case: When I use quick-reply to send a mail to someone of whom I have a trusted key, the mail still goes out unencrypted. I learned this by accidentally sending a mail in unencrypted form that I wanted to be encrypted. (Lucky enough, there was no personal or private data in that mail.)

Conversations could at the least inform users (maybe as part of its setup process) that it will never encrypt quick-replies. Even better, of course, would be extending the enigmail support to also cover sending :D

protz:
It uses the default preferences for your account.

There was an issue about that just a few weeks ago. Can you mark as a 
duplicate of #791?

Thanks,

Jonathan

RalfJung:
> It uses the default preferences for your account.

I have Enigmail configured to automatically encrypt mail from this account to verified keys. When I use "manual" reply, the mail gets encrypted without any further interaction from my side.
So, using the default does not work here, or I don't know what default it is using.

Should I still mark as dupe?

protz:
No, just cross-reference the issues. That's probably a wrong 
interpretation of "default" in the code.

Enigmail support is very primitive...

azrdev:
Using TB 38 and Conversations 2.10, Quick replies are not even signed, although they should be, as far as I've configured.

kehugter:
Same issue, quick replies aren't signed. I'm quite sure they were not long ago tho.

Standard8:
Looks like this was fixed by #1029 but didn't get closed when that was merged. If there's more missing, please feel free to file new issues.

Issue #845, opened 2014-08-13
Conversations 2.8.1 not working with TB 31 and Enigmail 1.7


dm74:
Hi, after upgrading to Conversations 2.8.1 I cannot see encrypted email... Conversations no longer asks me for the decryption password.
The only thing I can do is to open the message in the classic reader, which works as expected.

josephlhall:
Seeing this too, thanks much for the workaround that doesn't require a TB restart, @rgeddes

josephlhall:
Ok, the workaround is a bit different for me... here is what happens and what I do to get enigmail and conversations working together again:

1. Start TB.
2. It will crash, but allow you to open it anyway after clicking through a "cannot connect to the enigmail service" dialog.
3. Disable and then re-enable the conversations add-on (without restarting/quitting TB in between).
4. Then, quit and relaunch TB.

Happy to point to a crash report or whatever, if it helps.

MartinX3:
Same here:
FossaMail based on LTS Thunderbird v38.
There are some compatibility issues with conversations v2.10.2 and enigmail v1.9.6.1

Mails are sometimes not decrypted, instead, they are showing the encrypted stuff.
And the tags which are showing you if the mail is encrypted/signed are missing.
If i disable conversations, everything is working.

I already tried to deactivate the checkboxes in the wizard.
I think the problem is the conversation "view".

Standard8:
Looks like this was fixed by #1029 but didn't get closed when that was merged.

Issue #791, opened 2014-02-15
Enhance Enigmail interaction with more UI (especially quick reply)


klorydryk:
Hi,
When a conversation is encrypted with enigmail, if I use Conversation "quick response" plugin at some point, the email is sent without any encryption, with all conversation history.

It is very important to maintain the conversation protection. Add buttons to encrypt/sign ?

**Edit**: sub-issues in #711 #747 #640

bartpolot:
In the current version the behaviour I see is that 'quick reply' *always* sends the message unencrypted (no matter what the enigmail settings, per-recipient rules or conversation history) but does *not* include previous messages unencrypted, just the PGP ciphertext quoted.

protz:
Yes, this is a bug in the current version. Any help in fixing this would
be highly appreciated! :)

~ jonathan

bartpolot:
I wish I had the time... (don't we all?)

RalfJung:
Also see #1015, there's some trouble with Conversations fetching the default setting from Enigmail and applying it to quick replies.

Standard8:
Looks like this was fixed by #1029 but didn't get closed when that was merged. If there's more missing, please feel free to file new issues.

PR #1029, opened 2015-11-17
Enigmail 1.9 and quick reply support

### LINKED ISSUE #978: New API for interfacing Enigmail
We did a big refactoring of the Enigmail core. The consequences are that the JS-modules to include have changed and the various methods have moved to different modules.

Please  check the Enigmail source for details: https://sourceforge.net/p/enigmail/source/ci/master/tree/ or contact me if you need further infos.

shirosaki:
I fixed bugs with Enigmail 1.9 change and quick reply.
This works with Enigmail 1.8.
More tests would be required for quick reply support.
This will fix #978 #845 #791 #1015.

### DIFF
--- a/content/stub.compose-ui.js
+++ b/content/stub.compose-ui.js
@@ -744,7 +744,7 @@ ComposeSession.prototype = {
       for (let h of getHooks()) {
         try {
           if (typeof(h.onComposeSessionChanged) == "function")
-            h.onComposeSessionChanged(this, getMessageForQuickReply(), recipients);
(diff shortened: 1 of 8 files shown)


protz:
Fantastic! Thanks very much for the pull request.

Conversations is currently broken on trunk but as soon as I fix it I'll issue a new release with your patch in.

shirosaki:
Thunk you!
I tried to test this with Daily but got errors. I test with current release version 38.

protz:
Yes, I tried to fix Conversations to work with Daily, but there's 
something that changed (deeply), and I can no longer access the global 
variables from the HTML "multimessage" view from outside...

~ jonathan

PR #1030, opened 2015-11-18
Fix inline PGP verification and quick reply with Enigmail

shirosaki:
I fixed the following issue.

* inline PGP verification fails if a message has collapsed quoted text
* encrypt/sign mode changed manually is not always used on quick reply

And I did some refactoring and updated code from latest Enigmail source.

### DIFF
--- a/modules/plugins/enigmail.js
+++ b/modules/plugins/enigmail.js
@@ -208,8 +208,30 @@ if (hasEnigmail) {
   }, true);
 }
 
-function tryEnigmail(bodyElement, aMessage) {
-  if (bodyElement.textContent.indexOf("-----BEGIN PGP") < 0)
+function tryEnigmail(aDocument, aMessage, aMsgWindow) {
+  let bodyElement = aDocument.body;
+  let findStr = "-----BEGIN PGP";
+  let msgText = null;
+  let foundIndex = -1;
+  if (bodyElement.firstChild) {
+    let node = bodyElement.firstChild;
+    while (node) {
+      if (node.nodeName == "DIV") {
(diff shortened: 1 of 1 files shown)


modules/plugins/enigmail.js
```diff
@@ -228,45 +250,104 @@ function tryEnigmail(bodyElement, aMessage) {
 
   try {
     // extract text preceeding and/or following armored block
+    // strip "- show quoted text -" from body text
+    let NodeFilter = window.NodeFilter;
+    let treeWalker = aDocument.createTreeWalker(
+      bodyElement,
+      NodeFilter.SHOW_ELEMENT | NodeFilter.SHOW_TEXT,
+      { acceptNode: function (node) {
+          if (node.nodeType == 1) {
```
msprotz: use a constant here? i.e. `node.ELEMENT_NODE`

modules/plugins/enigmail.js
```diff
@@ -228,45 +250,104 @@ function tryEnigmail(bodyElement, aMessage) {
...
+          return NodeFilter.FILTER_ACCEPT;
+        }
+      }
+    );
+    let text = [];
+    while (treeWalker.nextNode())
+      text.push(treeWalker.currentNode.nodeValue);
+    msgText = text.join('');
+    msgText = msgText.replace(/\r\n/g, "\n");
+    msgText = msgText.replace(/\r/g, "\n");
```
msprotz: one regexp? `\r?\n/g`

protz:
That's a fantastic patch. Thanks!

I filed a couple of minor nits, so if you feel like fixing them, you can submit another commit / pull request, but otherwise, that fine, really.

PR #1031, opened 2015-11-18
Fixed JavaScript errors with Thunderbird trunk

shirosaki:
jQuery has been removed from Thunderbird.
We add jQuery 1.6.4 which is removed version.
https://bugzilla.mozilla.org/show_bug.cgi?id=1014609

Some global variables caused errors in stub.xhtml.
`StringBundle` in handlebars-wrapper.js.was undefined.
`let Log` in stub.xhtml caused redefinition error.


protz:
What can I say? This is awesome. Thanks!

PR #1033, opened 2015-11-19
Fix errors on quick reply with trunk

shirosaki:
We change the order of declaration to fix redeclaration and
reference errors.
`asToken` in stub.completion-ui.js is referenced in
stub.compose-ui.js. So we change to include stub.completion-ui.js
first.

Top-level const/let is lexical scope on trunk.
https://bugzilla.mozilla.org/show_bug.cgi?id=1209777


protz:
Thanks. I've tested this and everything works fine with Daily. This is great.

PR #1035, opened 2015-11-22
Show 'reply-to' in the message details

### LINKED ISSUE #1019: "Reply-To" is not shown
I have here a mail which has a `Reply-To` header. Conversations does not show that even when I enable "show details", unlike plain Thunderbird which does show this together with Cc, and so on.

RalfJung:
Fixes #1019

### DIFF
--- a/modules/message.js
+++ b/modules/message.js
@@ -1219,7 +1219,7 @@ Message.prototype = {
           value: sanitize(folderName(this._msgHdr.folder)[1]),
         });
         let interestingHeaders =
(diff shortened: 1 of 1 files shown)


protz:
I trust you that you actually checked that this works fine? :-)

Congratulations for your first pull request which, I'm sure, is only the first of a long series of pull requests to come.

RalfJung:
> I trust you that you actually checked that this works fine? :-)

I did, of course ;-)

Thanks for merging!

Issue #1020, opened 2015-10-18
Blank PGP-encrypted messages in conversation view


Manko10:
When I receive PGP-encrypted messages, I can read them without issues when I open them for the first time. But then when I select a different conversation and switch back to that conversation, all those messages are blank. Except for those that I wrote myself.

This is what it looks like:
![conversations-empty](https://cloud.githubusercontent.com/assets/911270/10563846/64ffdd3e-759b-11e5-865f-9d941a6795e9.png)

As you can see, the empty messages aren't even detected as being encrypted. In fact, when I turn on debugging and look into the log file, it says that it found three PGP messages, although there should be more (no errors are reported):

> 2015-10-18 13:13:13     Conversations.Modules.Enigmail  DEBUG   Found inline PGP
> 2015-10-18 13:13:13     Conversations.Modules.Enigmail  DEBUG   Found inline PGP
> 2015-10-18 13:13:13     Conversations.Modules.Enigmail  DEBUG   Found inline PGP

When I click "view using the classic reader", the message pane is still empty. Only when I right-click and say "Open Message in New Window" the contents are shown.

The only differences that I could find in the messages are that the ones that are shown correctly are all inline PGP whereas the others are multi-part MIME mails.
However, I tried sending myself multi-part messages, too and couldn't reproduce. So I examined the message sources and they look slightly different.

Enigmail creates multipart messages like this:

```
MIME-Version: 1.0
Content-Type: multipart/encrypted;
 protocol="application/pgp-encrypted";
 boundary="RIuCQL75xEBjjxeqDJ6WwlhdHq8fH4N4c"

This is an OpenPGP/MIME encrypted message (RFC 4880 and 3156)
--RIuCQL75xEBjjxeqDJ6WwlhdHq8fH4N4c
Content-Type: application/pgp-encrypted
Content-Description: PGP/MIME version identification

Version: 1

--RIuCQL75xEBjjxeqDJ6WwlhdHq8fH4N4c
Content-Type: application/octet-stream; name="encrypted.asc"
Content-Description: OpenPGP encrypted message
Content-Disposition: inline; filename="encrypted.asc"

-----BEGIN PGP MESSAGE-----
Version: GnuPG v2

...
-----END PGP MESSAGE-----

--RIuCQL75xEBjjxeqDJ6WwlhdHq8fH4N4c--
```

These are shown correctly.
The other messages I received which are only shown the first time I open them look like this:

```
MIME-Version: 1.0
[...]

PR #1037, opened 2015-11-23
Add notification bar UI and fix bugs with Enigmail

shirosaki:
* add notification bar UI to view signed status
* fix bugs with inline PGP on quick reply
* refactoring and trivial fix


protz:
Very cool! And congrats for finding the tricky bug with "... ? 2 : 1". I assume this fixes #1020?

shirosaki:
Thanks. But it seems #1020 is not fixed.
I can reproduce similar empty body bug.

PR #1038, opened 2015-11-24
Fix blank body of PGP/MIME messages

### LINKED ISSUE #1020: Blank PGP-encrypted messages in conversation view

shirosaki:
When I executed expand all messages, PGP/MIME messages in the
conversation was blank. If `lastMsgUri` is set, decrypting
messages other than `lastMsgUri` stops on Enigmail side.
It seems that this was changed on Enigmail at some point.
`lastMsgUri` is set from `Enigmail.msg.getCurrentMsgUriSpec`.
We remove these code to decrypt properly.
Security info display doesn't need `lastMsgUri` now.

This may fix #1020.

### DIFF
--- a/modules/message.js
+++ b/modules/message.js
@@ -667,7 +667,7 @@ Message.prototype = {
             self._conversation._htmlPane.scrollNodeIntoView(self._domNode);
             self.read = true;
           }
-        }, (self.needsLateAttachments ? 2 : 1));
+        }, 1);
         self.toggle();
       }, false);
 
--- a/modules/plugins/enigmail.js
+++ b/modules/plugins/enigmail.js
@@ -422,7 +422,6 @@ function tryEnigmail(aDocument, aMessage, aMsgWindow) {
 function verifyAttachments(aMessage) {
   let { _attachments: attachments, _uri: uri, contentType: contentType } = aMessage;
   let w = topMail3Pane(aMessage);
-  w.Enigmail.msg.getCurrentMsgUriSpec = function () uri;
   if ((contentType+"").search(/^multipart\/signed(;|$)/i) == 0) {
     w.Enigmail.msg.messageDecryptCb(null, true, {
       headers: {'content-type': contentType },
@@ -601,6 +600,8 @@ let enigmailHook = {
       if (hasSig)
         aMessage._domNode.classList.add("signed");
 
+      // Current message uri should be blank to decrypt all PGP/MIME messages.
+      w.Enigmail.msg.getCurrentMsgUriSpec = function () { return ""; }
       verifyAttachments(aMessage);
       prepareForShowHdrIcons(aMessage);
       patchForShowSecurityInfo(w);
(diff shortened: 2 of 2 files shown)


shirosaki:
Sorry. I misunderstood signal count. Signal was not called if needsLateAttachments.
I saw signal call error, but It would be due to other reason.

protz:
Hey Hiroshi,

Thanks for taking care of all these pesky issues. I must admit that I don't have time to investigate all these issues related to the interaction with Enigmail, so your help is really invaluable!

PR #1040, opened 2015-11-26
Improve default encrypt/sign status on quick reply

shirosaki:
We use per-recipient rules to set default encrypt/sign status on
quick reply. This follows Enigmail behavior. Code is based on
enigmailMsgComposeOverlay.js in Enigmail.
`EnigmailConstants` is not available in Enigmail 1.8.2. We fixed
to work with 1.8.2.

### DIFF
--- a/modules/plugins/enigmail.js
+++ b/modules/plugins/enigmail.js
@@ -151,6 +151,9 @@ if (hasEnigmail) {
     hasEnigmail = false;
     Log.debug("Enigmail script doesn't seem to be loaded. Error: " + e);
   }
+  if (!global.EnigmailConstants) {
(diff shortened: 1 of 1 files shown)


protz:
Thanks again. Please let me know if I should put something in the release notes for the next version (such as: a minimum version of Enigmail required).

shirosaki:
Thanks for merge.
I test it with Enigamil v1.8.2 (latest release version)  and nightly build 1.9a1pre.
I'm not sure about before v1.8.2. Minimum version of Enigmail would be v1.8.2.

protz:
Excellent. I'll mention that in the release notes for the next version.

PR #1046, opened 2015-12-24
Update 'more' link and tooltip UI

shirosaki:
Implement 'more' link and tooltip UI based on the mockups in #967.

I tried to implement the UI partially. What do you think?

ui


msprotz:
Hi Hiroshi,

I just took a look at this. I think it's a really good start! Some remarks:
- the vertical dots are not really aligned on my computer; 
image you probably need to use some different CSS to achieve the desired effect
- reply became "Reply"; the case is not consistent with the rest of the UI
- the images in the menu are not aligned vertically (`vertical-align: middle` ?) 
image
- the images in the menu don't look good on my hidpi screen; do you have an SVG source for them or a set of alternative, 2x resolution images?

Thanks and a merry christmas!

Jonathan

shirosaki:
Thank you for review!

I fixed vertical align issues to use `vertical-align: middle`.
I changed the case small as the rest of the UI.
I don't have SVG or high resolution icons. I made icon images from the jpeg mockups of @elioqoshi by copy and modification. Does @elioqoshi have it?

shirosaki:
I found that the icons are from http://www.flaticon.com/authors/google.
So I've replaced icons with SVG.

protz:
@shirosaki : I haven't forgotten about this pull request, I'm just really busy right now. I'll do a proper review in early February once I'm back in Seattle.

Thanks for understanding!

protz:
(But overall I think it's good and I'll end up integrating it. Feel free to keep working on this if you feel so inclined...)

xi:
This is probably dependent on the theme, but for me the text changes color (from dark to light) on hover. It would be nice if the icon always had the same color as the text. This should be possible using icon fonts.

xi:
These changes have been incorporated into #1141, so this pull request can be closed now.

xi:
Thanks to @shirosaki for the research and inspiration!

Issue #1056, opened 2016-01-29
Add-on broken in latest TB (46.0) betas


darrenkenny:
It looks like there have been some JS changes in TB 46 beta that is causing the loading
of the conversations add-on to fail.

For example the Array comprehension doesn't seem to work any more, etc.

Any chance of the add-on being updated to deal with these issues?

protz:
Please install 
http://jonathan.xulforum.org/files/gcv-nightlies/201511251059-master.xpi 
and close this issue if it fixes your problem.

Thanks,

Jonathan

darrenkenny:
Thanks for the quick response!

Unfortunately, that's the version that I'm running - always have tried to run the latest that you
have available there.

So, here are the exact errors that I'm seeing when I enable the add-on after restarting TB with it
disabled:

`Timestamp: 29/01/2016 12:11:34
Warning: expression closures are deprecated
Source File: resource://conversations/modules/prefs.js
Line: 80, Column: 22
Source Code:
  split: function (s) Array.map(s.split(","), String.trim).filter(String.trim).map`

`1454069494931	addons.xpi	WARN	Exception running bootstrap method startup on gconversation@xulforum.org: SyntaxError: missing ] after element list (resource://conversations/modules/prefs.js:115:21) JS Stack trace: @log.js:3:1 < startup@resource://gre/modules/addons/XPIProvider.jsm -> jar:file:///D:/Users/darrenk/AppData/Roaming/Thunderbird/Profiles/t4tlfmds.default/extensions/gconversation@bootstrap.js:170:3 < this.XPIProvider.callBootstrapMethod@XPIProvider.jsm:4658:9 < this.XPIProvider.updateAddonDisabledState@XPIProvider.jsm:4785:13 < AddonWrapper.prototype.userDisabled@XPIProvider.jsm:7062:9 < set_userDisabled@extensions.xml:1072:11 < oncommand@about:addons:1:1`

`Warning: attempting to write 11491 bytes to preference extensions.xpiState. This is bad for general performance and memory usage. Such an amount of data should rather be written to an external file.`

Thanks,

Darren

protz:
Are you on trunk ? Beta is 44 right now

darrenkenny:
Apologies, it is EarlyBird, so yes trunk. Sorry for the confusion.

protz:
[...]

PR #1058, opened 2016-02-06
Replace the deprecated uses of expression closures with arrow functio…

Dexterp37:
…ns. Also changes for-of to Array.map where needed.

Along with the thunderbird-stdlib changes, this fixes the addon to work with the current Thunderbird Daily and takes care of issue #1056 :-)

### DIFF
--- a/content/stub.compose-ui.js
+++ b/content/stub.compose-ui.js
@@ -123,7 +123,7 @@ function registerQuickReply() {
           obj.onDraftChanged(aTopic);
         }
         // While we're at it, cleanup...
-        listeners = listeners.filter(function (x) x.get());
+        listeners = listeners.filter(x => x.get());
         mainWindow.Conversations.draftListeners[id] = listeners;
       } catch (e) {
         // See comment above
@@ -187,7 +187,7 @@ function newComposeSessionByDraftIf() {
   SimpleStorage.spin(function () {
     let r = yield ss.get(id);
     if (r) {
-      gComposeSession = createComposeSession(function (x) x.draft(r));
+      gComposeSession = createComposeSession(x => x.draft(r));
       startedEditing(true);
       revealCompositionFields();
       showQuickReply.call($(".quickReply li.reply"));
@@ -205,9 +205,7 @@ function newComposeSessionByClick(type) {
   Log.debug("Setting up the initial quick reply compose parameters...");
   let messages = Conversations.currentConversation.messages;
   try {
-    gComposeSession = createComposeSession(function (x)
-      x.reply(getMessageForQuickReply(), type)
-    );
+    gComposeSession = createComposeSession(x => x.reply(getMessageForQuickReply(), type));
     // This could probably be refined, like only considering we started editing
     // if we modified the body and/or the composition fields.
     startedEditing(true);
@@ -596,7 +594,7 @@ ComposeSession.prototype = {
       if (!aDefault) // "" evaluates to false
         return aList;
       for (let email of aDefault.split(/,/)) {
-        if (!aList.some(function (x) x.email == email)) {
+        if (!aList.some(x => x.email == email)) {
           aList.push(asToken(null, null, email, null));
         }
       }
@@ -773,8 +771,7 @@ ComposeSession.prototype = {
     else
       compType =  Ci.nsIMsgCompType.ReplyAll; // ReplyAll, Reply... ends up the same
 
-    let [to, cc, bcc] = ["to", "cc", "bcc"].map(function (x)
-      JSON.parse($("#"+x).val()));
+    let [to, cc, bcc] = ["to", "cc", "bcc"].map(x => JSON.parse($("#"+x).val()));
 
     let sendStatus = { };
     for (let priority of ["_early", "", "_canceled"]) {
@@ -929,7 +926,7 @@ AttachmentList.prototype = {
     });
     line.find(".removeAttachmentLink").click(function () {
       line.remove();
-      self._attachments = self._attachments.filter(function (x) x != msgAttachment);
+      self._attachments = self._attachments.filter(x => x != msgAttachment);
     });
     line.appendTo($(".quickReplyAttachments"));
   },
@@ -958,11 +955,13 @@ AttachmentList.prototype = {
   },
 
   save: function () {
-    return [{
-      name: x.name,
-      size: x.size,
-      url: x.url,
-    } for (x of this._attachments)];
+    return this._attachments.map(x => {
+      return {
+        name: x.name,
+        size: x.size,
+        url: x.url,
+      };
+    });
   },
 
   get attachments () {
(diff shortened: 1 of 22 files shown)


modules/assistant.js
```diff
@@ -104,12 +104,12 @@ MixIn(PrefCustomization, SimpleCustomization.prototype);
 
 
 function MultipleCustomization(aParams) {
-  this.customizations = [new PrefCustomization(p) for (p of aParams)];
+  this.customizations = !!aParams ? aParams.map(p => new PrefCustomization(p)) : [];
```
benjaoming: This looks like it might have fixed a bug? Should the same pattern be used other places or, or was this a special case?

benjaoming:
Great work! As it touches a huge portion of code, recommending to merge this ASAP!

modules/assistant.js
Dexterp37: That's a good point! By having a deeper look at the code, it seems like this is only used at modules/assistant.js:133, so it might be safe to just change that to `this.customizations = aParams.map(p => new PrefCustomization(p));`.

That approach is safer (but one could argue that it depends on how it's used) if the argument can be null or undefined, which doesn't seem the case here. My bad :-)

content/stub.compose-ui.js
```diff
@@ -958,11 +955,13 @@ AttachmentList.prototype = {
...
-      size: x.size,
-      url: x.url,
-    } for (x of this._attachments)];
+    return this._attachments.map(x => {
+      return {
+        name: x.name,
+        size: x.size,
+        url: x.url,
+      };
+    });
```
msprotz: `x => ({ name: x.name, ... })` here?

content/stub.xhtml
```diff
@@ -479,9 +476,9 @@
             window.frameElement.setAttribute("tooltip", "aHTMLTooltip");
           let mainWindow = topMail3Pane(window);
           let willExpand = parseInt(params.willExpand);
-          let msgHdrs = [msgUriToMsgHdr(x) for (x of params.urls.split(","))];
-          msgHdrs = msgHdrs.filter(function (x) x != null);
-          msgHdrs = msgHdrs.filter(function (x) x.messageId);
+          let msgHdrs = params.urls.split(",").map(x => msgUriToMsgHdr(x));
+          msgHdrs = msgHdrs.filter(x => x != null);
+          msgHdrs = msgHdrs.filter(x => x.messageId);
```
msprotz: nit: you could clean this up and do; `params.url.split(...).filter(x => x != null && x.messageId)`

modules/assistant.js
```diff
@@ -125,7 +125,7 @@ MultipleCustomization.prototype = {
 //  "Illegal operation on WrappedNative prototype object"  nsresult:
 //  "0x8057000c (NS_ERROR_XPC_BAD_OP_ON_WN_PROTO)"
 // So we do a round of eta-expansion.
-let eid = function (id) getMail3Pane().document.getElementById(id);
+let eid = (id) => getMail3Pane().document.getElementById(id);
```
msprotz: why `(id)` and not `id`?

modules/conversation.js
```diff
@@ -388,9 +389,8 @@ Conversation.prototype = {
   _selectionChanged: function _Conversation_selectionChanged () {
     let gFolderDisplay = topMail3Pane(this).gFolderDisplay;
     let messageIds = this._initialSet.map(x => x.messageId);
-    return
```
msprotz: oh you probably fixed a bug here!

modules/message.js
```diff
@@ -236,7 +236,7 @@ KeyListener.prototype = {
   findMsgNode: function findMsgNode(msgNode) {
     let msgNodes = this.message._domNode.ownerDocument
       .getElementsByClassName(Message.prototype.cssClass);
-    msgNodes = [x for (x of msgNodes)];
+    msgNodes = Array.from(msgNodes);
```
msprotz: nice

modules/message.js
```diff
@@ -744,14 +744,18 @@ Message.prototype = {
     // We run below code only for the first time after messages selected.
     Log.debug("A message is selected: " + this._uri);
     this._selected = true;
-    [message._selected = false
-      for ({ message } of this._conversation.messages)
-      if (message != this)];
+    for (let message of this._conversation.messages) {
```
msprotz: I think this one should be:

```
for let { message } of ...
   if (messge != this)
```

msprotz:
Excellent pull request. Thanks for taking care of this!

I'll let you address the nits then I'll do the merge.

content/stub.xhtml
Dexterp37: I assumed you didn't mean to remove the |map| part?

Dexterp37:
Thank YOU for Thunderbird Conversations :-)

content/stub.xhtml
protz: No :)

protz:
Have you tested this? It doesn't fix anything for me I'm still getting errors...?

2016-02-11 12:31:08	Conversations.MonkeyPatch	ERROR	TypeError: this.watchers is undefined

Dexterp37:
Yes, I tested this with Thunderbird Daily 47.0a1 (2016-02-05), and I'm able to see the conversations threads and to interact with them (expand the thread, delete the whole stack of conversations, remove a single email).

Maybe I missed some other required change, as the regexp I used to list the changes might have missed some. How are you triggering that error?

msprotz:
It's at start-up time...

protz:
Thanks. I've started looking into locally, and you need this to start with:

```
diff --git a/modules/prefs.js b/modules/prefs.js
index 11767b8..94a2ba6 100644
--- a/modules/prefs.js
+++ b/modules/prefs.js
@@ -79,7 +79,9 @@ PrefManager.prototype = {
 
   split: s => Array.map(s.split(","), String.trim).filter(String.trim).map(String.toLowerCase),
 
-  watch: watcher => this.watchers.push(watcher),
+  watch: function (watcher) {
+    this.watchers.push(watcher)
+  },
 
   register: function mpo_register (observer) {
     prefsService.QueryInterface(Ci.nsIPrefBranch);
```

then there's more occurrences of the old syntax in `content/stub.xhtml`

protz:
(Capturing `this` lexically is incorrect because it's a "prototype method", meaning you want the dynamic `this` that is bound to the receiver of the method call.)

Dexterp37:
(Thanks, I eventually figured that out :-) )

This should be it, I took care of those remaining issues. Nothing seems to be broken.

callahad:
So far, the `arrow_funcs` branches in @Dexterp37's forks of this repo and the stdlib repo seem to be working well for me on Earlybird, though I did have to change the commit that the submodule points to for `modules/stdlib`.

Thanks, @Dexterp37 !

protz:
This works fine. Thanks!

protz:
Good catch about the gitmodule. I fixed it with an extra commit.

@Dexterp37 that was great work! One outstanding issue is the "show remote content" thing that no longer works on trunk. If you are motivated, I'd be more than happy to provide guidance on how to fix this :)

protz:
Timestamp: 02/17/2016 02:46:52 PM
Error: TypeError: entries(...).map is not a function
Source File: chrome://conversations/content/stub.completion-ui.js
Line: 96

@Dexterp37 another pull request perhaps?

protz:
:)

Dexterp37:
Thanks :-) Sure, I'll check that over the weekend!

PR #1060, opened 2016-02-20
Fix suggestions when adding a new contact to an email.

Dexterp37:
This fixes:

Timestamp: 02/17/2016 02:46:52 PM
Error: TypeError: entries(...).map is not a function
Source File: chrome://conversations/content/stub.completion-ui.js
Line: 96

And allows to see the suggested contacts when adding a new contact to an email!


msprotz:
Cool! Does it also fix the "quick reply"? (Replying inline to an email)

Dexterp37:
Adding contacts in the quick reply now works. Or did you mean something else?

protz:
That seems to fix pretty much everything. Thank you so much!

Issue #1091, opened 2016-05-24
Remember opened tabs with conversations


diegogurpegui:
I close Thunderbird with some tabs opened (Conversation tabs), but then when I open Thunderbird again, none of those remembered tabs open correctly. They keep with the title "Conversation Reader", the conversation never loads and the tab content is just blank.

This used to work great since the beginning. However since the last update of Thunderbird and/or Conversations (not sure which) it stopped working.
If I open a new conversation from the Inbox, it opens perfectly.

diegogurpegui:
If it's any help: I use a Master Password

Linuspogo:
I have the exact same issue, on both the stable and the beta version of Thunderbird. I don't use a Master Password.

francky37:
Same issue here, "Loading.." is displayed at the top, but nothing happens

jfrancin:
Exact same symptoms, using Thunderbird 45.1.1 and Thunderbird Conversations 2.11.4.  This used to work.

theWoosh:
I too have this problem - it has happened intermittently (and annoyingly) in recent months but now it happens every time I restart thunderbird... does anyone have any clues? I am using thunderbird 45.1.1 on Windows 8.1...

theWoosh:
how do we escalate this problem?

xi:
I did some debugging and found that it works correctly if I remove the `_selectionChanged()` check from [`_Conversation_outputMessages()`](https://github.com/protz/thunderbird-conversations/blob/9704a57/modules/conversation.js#L815). This seems to be very old code, so I do not really dare to touch it. But maybe it can just be removed.

francky37:
To xi,
Thank you very much, I removed the lines and it works perfectly now.

theWoosh:
hi xi - where do I find this file to modify? I have had this problem occasionally for more than a year, but in the last weeks it has become chronic...

xi:
**Warning**: I do not recommend to do this. It should rather be fixed upstream. Do this at your own risk.

Go to your profile's extension folder (mine is at `~/.thunderbird/{profilename}/extensions/`). There should be a file called `gconversation@xulforum.org.xpi`. Create a folder called `gconversation@xulforum.org` (same name without extension), and unpack the file to that folder. The file you need to edit is `modules/conversation.js`.

tvankess:
[...]

Issue #1090, opened 2016-05-22
"Allow remote content from..." problems


gregorko:
When I activate the plugin the remote content privacy options menu show strange entries:

![clipboard-2](https://cloud.githubusercontent.com/assets/647058/15455250/0ba216c6-204f-11e6-843f-6024776e6d70.jpg)

When I disable the plugin the menu shows correctly:

![clipboard-1](https://cloud.githubusercontent.com/assets/647058/15455260/2c8504de-204f-11e6-8644-ffc933fd1dfe.jpg)

protz:
Yes... there was a corresponding discussion somewhere in bugzilla and then I couldn't understand what needed to change on my side to fix this. Pull request welcome.

protz:
https://bugzilla.mozilla.org/show_bug.cgi?id=1266356

Standard8:
The UI is much simpler now, and this doesn't happen.

PR #1123, opened 2016-11-08
Fixing issue #1091

daesters:
Fixes issue #1091


protz:
Sorry -- I just now saw this.

Before merging, I'd like to understand:
- is the function gone?
- is the function no longer relevant, or does it throw an exception?
- isn't it better to wrap it in a try-catch block?
- is this codepath exercised only when in a tab?

Thanks,

Jonathan

xi:
This may come from my comment at https://github.com/protz/thunderbird-conversations/issues/1091#issuecomment-227230417. Some people reported that it helps, but I strongly advise against merging it without a better understanding of the problem.

That said, #1091 is a real dealbreaker and I would really like if it was fixed soon.

daesters:
@protz Sorry, I didn't reply earlier.
As @xi says, it was his comment which pointed to the solution. I don't know the details about the function, whether it is obsolet now or an exception would be better. I'm not yet deep into addons programmming of thunderbird.
I just implemented it, because, #1090  "is a real dealbreaker",as @xi rightly says

protz:
I've investigated and pushed a proper fix. Thanks! (Pre-release build at https://jonathan.protzenko.fr/thunderbird-conversations/)

Standard8:
For reference, the proper fix was https://github.com/protz/thunderbird-conversations/commit/f2252db9bda1c2d19f5e1a47506302fae9867056

PR #1126, opened 2016-11-23
replaced all references to the old name

SimJoSt:
replaced all `GMail-Conversation-View` strings by `thunderbird-conversations`

- helps to avoid confusion in addon detail view
- links don't need redirect anmyore


protz:
Thanks! That's very useful

PR #1133, opened 2016-12-31
Fix installation hanging

protz:


protz:
Thanks Paul

pianoslum:
You're welcome!

But does this need more work (I just deleted the faulty lines)?
Do you have any clue ehy
`mainWindow.MsgSortThreaded();` stopped working?

And should we replace "beetweenColumn" by "correspondentsColumn" or something or is this obsolete now?

msprotz:
Yes, I think we can improve this.
- We need to find the new equivalent of MsgSortThreaded... this is not too hard, just need to figure out what happens when one clicks View > Sort By > Threaded (there's a command dispatched, and then searching DXR for the command name should help).
- Yes, we should show the correspondents column and revert it if possible.

PR #1137, opened 2017-01-07
Update jQuery

pianoslum:
I noticed the following warning

Zeitstempel: 07.01.2017 09:52:23
Warnung: getPreventDefault() sollte nicht mehr verwendet werden. Verwenden Sie stattdessen defaultPrevented. [... is deprecated, use ... instead)
Quelldatei: chrome://conversations/content/js/jquery-1.6.4.min.js
Zeile: 3

Our jQuery-version is from 2011, so I updated it to the latest 1.x branch. I only did some short testing, and everything seems ok. But some more testing would be good before you merge it!


protz:
Looks good to me, thanks!

PR #1138, opened 2017-01-07
First steps for design update

xi:
This implements some first steps for implementing the new design proposed in #967, namely:

- increase message header height
- remove rounded corners
- change shadows

@protz I do have some more in the pipeline. I am not sure whether you prefer many small or one big pull request. As I am not sure how much time I have in the next weeks I would prefer to do this in many small steps. This would reduce the threat of merge conflicts.


protz:
Thanks. I've branched out 2.12+ right before the merge so that I can issue a bugfix release without the graphic changes.

PR #1139, opened 2017-01-09
refactor CSS

xi:
As a further step towards implementing the design from #967 I did some general CSS refactoring. This should only have minimal UI changes.


protz:
How can I say no to a pull request that removes a lot of code? Thanks!

protz:
I noticed the following:

image

Is it intentional that there's a double border between messages, or is it a work in progress?

xi:
> Is it intentional that there's a double border between messages, or is it a work in progress?

Kind of both: In the [design](https://github.com/protz/thunderbird-conversations/issues/967#issuecomment-147186174) there are 1px gaps between the messages. So this is where that comes from.

At the same time, the design has [issues with contrast](https://github.com/protz/thunderbird-conversations/issues/967#issuecomment-270784630). That is why I did not remove the borders yet.

PR #1141, opened 2017-01-12
Use material design icons

xi:
Another step towards #967.

I used an inline-svg method. This is a bit ugly in the html but allows to easily style the icons in CSS.

Note that some icons are signigicantly different, e.g. the ones in the conversation header.


protz:
Ok it seems like now the junk button is always shown, whereas previously it was conditionally shown depending on whether the message had the junkable class or not. Could it be that's something that your patch changed? 

Also, it looks like the animation for the quick reply fields is now fixed. That's great!

xi:
> Ok it seems like now the junk button is always shown, whereas previously it was conditionally shown depending on whether the message had the junkable class or not. Could it be that's something that your patch changed?

I cannot reproduce that issue. Any more information?

protz:
Normally the junk button only appears when a single message is selected in the message list.

protz:
This is `_updateConversationButtons`... but not sure whether it's your patch that caused the issue or not

protz:
Just ran a quick regression -- the bug isn't currently on origin/master 
so I'm afraid this pull request introduces it.

xi:
Works fine in my setup though, so it is hard for me to debug.

I did some refactoring of the relevant code in 2e327c0. Maybe that helps.

protz:
I think you need to select a collapsed thread for the junk button to 
disappear...

protz:
image

xi:
What version of thunderbird are you using? Have you tried restarting thunderbird?

protz:
52.0a2 -- yes, I have restarted Thunderbird.

TypeError: markReadButton is null[Learn More] conversation.js:1049:7

This is the code:

1044     // Make sure the toggle read/unread button is in the right state
1045     let markReadButton = 
this._htmlPane.document.querySelector("span.read");
1046     if (this.messages.some(x => !x.message.read))
1047       markReadButton.classList.add("unread");
1048     else
1049       markReadButton.classList.remove("unread");

I suspect you changed these classes and did not update the corresponding 
selectors in the JS code.

xi:
I fixed that in 2e327c0.

protz:
Oh, excellent! Looks good to me, pushed to master.

protz:
Thanks very much, this is starting to look really good.

PR #1142, opened 2017-01-15
Some small UI changes

xi:
I have some more changes for #967 locally, but I am not pleased with most of them yet. So this is a small collection of bits and bobs that can already be merged.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -18,6 +18,10 @@ Jonathan Protzenko.
 Portions created by the Initial Developer are Copyright (C) 2010
 the Initial Developer. All Rights Reserved.
 
+Icons under skin/material-icons.svg made by Google (https://material.io/icons/)
+are licensed by Apache License Version 2.0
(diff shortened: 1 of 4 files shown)


protz:
Looks good, thanks!

xi:
That was fast!

protz:
I need to keep up with your pull requests and make sure they're treated 
with the consideration they deserve! ;-)

protz:
I think this pull request or another one broke the contact tooltip...

protz:
That is, after hitting "show details"...

xi:
> I think this pull request or another one broke the contact tooltip...

What exactly seems to be broken? For me it seems to work just fine.

protz:
image

When I hover over the contacts in that view, I no longer get a contact tooltip... but perhaps it was already the case before?

xi:
> but perhaps it was already the case before?

Yes, I think so. II checked it on 00f9ec002088780e910668dceea7a352488c91e6 and the behavior was the same as on master.

PR #1143, opened 2017-01-15
use nativehyperlinktext color

xi:
As mentioned in https://github.com/protz/thunderbird-conversations/issues/967#issuecomment-272686468, there is a problem with link colors if the chrome and composition colors are significantly different. This change might be a way to fix that: For the chrome, use `-moz-nativehyperlinktext` instead of `-moz-HyperlinkText`. 

Note that the value for `-moz-HyperlinkText` can be controlled from the UI (Preferences -> Display -> Formatting -> Colors). The matching background color is however not exposed to CSS (As far as I know).

I am not completely sure if this is a good idea. It should be tested with some different setups.

### DIFF
--- a/skin/conversation.css
+++ b/skin/conversation.css
@@ -87,7 +87,7 @@ a img {
 
 .link {
   text-decoration: none;
-  color: -moz-HyperlinkText;
+  color: -moz-nativehyperlinktext;
   cursor: pointer;
 }
 
--- a/skin/quickreply.css
+++ b/skin/quickreply.css
@@ -8,7 +8,7 @@
 
 .quickReply a {
   text-decoration: none;
-  color: -moz-HyperlinkText;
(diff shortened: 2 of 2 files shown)


protz:
image

whereas, before:

image

protz:
Would be good to have screenshots with default gtk theme, and windows also -- I will not be able to provide that until a couple weeks when I'm back home (currently traveling).

xi:
I may be able to produce these screenshots, but it will take a while. I guess this is not urgent and also does not block any other changes.

elioqoshi:
Can we get rid of the small white space between 2 blocks?. Ideally there should be no space.

protz:
Merged this -- I'll leave it up to you to fix the small gap in-between the messages in a subsequent pull request. Thanks for the very valuable work!

PR #1145, opened 2017-01-20
[Redesign] Use Icon Font instead of inline SVG

fllp:
### Rationale

I discovered that using an icon font in Thunderbird works very well and is quite easy (see the [officical guide](https://google.github.io/material-design-icons/#using-the-icons-in-html)).  So this PR replaces the inline, 'one-in-all' SVG from f9b3471ec0435670987b7132a1754263ee442d39 with the font for Material Icons.  
This might add some 'overhead' since only a fragment of the icons are used. But it should offer more flexibility when updating or changing icons while a the same time keeping the ability to control color and size via CSS.

### Changes

- remove all the inline SVGs and use `<i>`-Tags with special classes  
and reference icons by using their ligature handle.
- replace tag icon for encryption to vpn key (was enhanced encryption)
- use star border instead of star for starring
- also clean up XML code a bit

### Issues

- [x] ~~the icons of pgp/mime tags are off~~  
- position of expand/collapse arrows is a little too low  
expand-glitch
- the icon for starring only toggles color  
(requires a different icon tag and probably some scripting)

### Screen

small-changes


fllp:
@protz, @xi what do you think?

msprotz:
This looks good, except the hollowed out star makes it look like it's not really starred... but maybe I'm just used to the old design.

fllp:
Yes, you're right.  It looks a bit odd
In the original design the hollow star used for unmarked messages only. I just could'nt find a quick way to toggle the full star icon properly and left it there as a reminder of sorts.  

In the meantime i managed to fix the alignment problem for pgp/mime tags.

msprotz:
Shall I leave this open while you try to iterate over the design for the star?

fllp:
Either that or i could change the icon to the solid version (its not
even an one liner). I woulndt look off and implementing the icon toggle
according to the design proposal could then be done in another branch/PR.
Since having the icon font available should simplifiy and speed up
adding more icons, let's hear @xi's opinion first.

xi:
When working on #1141 I first used an icon font and later switched to inline SVG. I used this approach after reading about the benefits of that approach (e.g. [github has recently switched to inline SVG](https://github.com/blog/2112-delivering-octicons-with-svg); [same for lonelyplanet](http://ianfeather.co.uk/ten-reasons-we-switched-from-an-icon-font-to-svg/)).

To be honest, I am not totally convinced of either approach. Here is an incomplete list of cons for each one:

- icon fonts
  - antialising optimized for text may result in sub-optimal quality
  - positioning via CSS is optimized for text, so it sometimes feels hacky
  - there are accessibility issues
  - seems like it is deprecated in favour of inline SVG in the long term
- inline SVG
  - is currently not as well known by developers
  - requires verbose markup

> Since having the icon font available should simplify and speed up adding more icons

I would disagree with that argument. We can just as well include the full icon set in the 'one-in-all' SVG file (I propably should have done that in the first place). So this flexibility is not related to the question of inline SVG vs. icon font. It works with either approach.

I really don't know which approach is preferable. I have nothing to say against this pull request. We should just make an informed decision and stick to one approach after that.

fllp:
Ok, this is why i wanted to hear your opinion first, @xi. I read the
post over at GitHub and have to admit that i didn't know about these
issues. It seems not all that glitters is gold, huh?
On the other hand it just works. But does that outweigh the issues?

And since TB has enough issues with HDPI already using an icon font with
its rendering quirks might worsen that.
Regarding accesibility (e.g. screen readers) i figured, aria-labels
might help a little. Though i think that wont be of much help.
I'd like to take more time to investigate and discuss, but currently i
lack the time.

@protz: so, i guess this should stay open for discussion now. unless, of
course, you already know you dont want icon fonts.

protz:
Re-reading the arguments (and the two linked articles), it seems to me like inline svg is superior, especially for rendering.

Would you be ok reverting to inline svg? Then, I'd be delighted to merge this pull request. Thanks!

fllp:
I agree, SVG seems to be the better choice here.
I can try to reverse the changes this week, but my time is somewhat
limited right now.
Anyway, since most of my changes were replacing the existing SVG spans
with icon fonts, i'm not sure if its worth merging.
Once i revert the icon font replacements, there will be just 2 changed
icons and some code clean-up left, i think.


Am 12.02.2017 um 01:57 schrieb Jonathan Protzenko:

protz:
Still happy to take in whatever cleanups you may have.

xi:
About the star icon: It is probably easier to use SVG styling (`stroke`/`fill`) instead of a separate icon.

fllp:
I'm closing this. The relevant changes aside icon fonts (smal code cleanups) will be moved to a separate branch/PR. Should be nicer than reverting all the commits with the icon font changes.

PR #1146, opened 2017-01-22
Followup to #1133

pianoslum:
MsgSortThreaded should be ok now, as the fix on TB side is included in both ESR and "normal" release.

Regarding the correspondent column: What's the advantage over the sender column? For me it looks quite the same, save for the little arrow before.

### DIFF
--- a/modules/assistant.js
+++ b/modules/assistant.js
@@ -199,7 +199,7 @@ let Customizations = {
       let moveOn = function () {
         let tabmail = mainWindow.document.getElementById("tabmail");
         tabmail.switchToTab(0);
-        //mainWindow.MsgSortThreaded();
+        mainWindow.MsgSortThreaded();
         /**
          * We don't know how to revert these, so forget about it for now.
          */
--- a/modules/assistant.js
+++ b/modules/assistant.js
@@ -210,6 +210,7 @@ let Customizations = {
         state.senderCol = eid("senderCol").getAttribute("hidden");
         eid("unreadCol").setAttribute("hidden", "false");
         eid("senderCol").setAttribute("hidden", "true");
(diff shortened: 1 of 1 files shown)


msprotz:
The correspondents column is more in the spirit of the conversation view in that it shows you all the people involved in a conversation... thought that'd be a nice customization to include.

Years ago, that column didn't exist, and the addon used to provide a "between" column that was a substandard version of what is currently the correspondents column.

What are your thoughts on that customization?

pianoslum:
Actually it doesn't work for me... Only one correspondent is shown at the time even though several are part of the conversation. Is it because TB wouldn't have them grouped together without Conversations?

Anyway, if it works for others it's ok by me - you can change it with one click ;-)

protz:
Right... I think they changed that, it didn't use to be the behavior.

Do you know when that happened?

If I had to speculate, I would say they caved in due to some pressure 
from old-timers who couldn't stand their favorite email client deviating 
from the 1992 behavior.

If we can override that behavior and make it a true correspondents 
column, then we should do that. Otherwise, sounds like we need to 
reinstate the old "between" column...

pianoslum:
It seems that the correspondents-column was never intended to be smart:
From https://bugzilla.mozilla.org/show_bug.cgi?id=1265446:

> The Correspondents field works like this:

> * Same as "From" for received messages
> * Same as "Recipients" for sent messages

Pretty lame, huh? So I'm for resurrecting the 'between'-column. As Conversation does better threading than TB, the info couldn't have been complete anyway...

msprotz:
yes but you need to revert my earlier changes in monkeypatch.js that removed the definition of the between column... it was entirely implemented by me in JS

pianoslum:
See #1148

Issue #622, opened 2012-07-12
Conversation View 2.4 break IMAP Folder localisation DE


BlueCase:
Hi,
since I updated to 2.4 I have the problem that my mail function folders (Inbox, Sent, Template....) are in English and not German. 
Switch back to 2.3.2 solved this problem.

regards,
Marco

PR #1148, opened 2017-01-28
Reintroduce between column

pianoslum:
This should reintroduce the 'between'-column, which is indeed better than the correspondents-column btw.

Incidentally, I updated stdlib

### DIFF
--- a/modules/assistant.js
+++ b/modules/assistant.js
@@ -199,7 +199,7 @@ let Customizations = {
       let moveOn = function () {
         let tabmail = mainWindow.document.getElementById("tabmail");
         tabmail.switchToTab(0);
-        //mainWindow.MsgSortThreaded();
+        mainWindow.MsgSortThreaded();
         /**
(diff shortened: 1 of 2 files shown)


pianoslum:
I forgot: The between-column's label is displayed in English for me, although my lang is set to German. Did I forget to revert something or is this expected?

msprotz:
https://github.com/protz/thunderbird-conversations/issues/622 ?

pianoslum:
No, if so, more names belonging to TB itself would be english.
I just saw that "Reply" etc. on Conversations side is still English - I recall that you said that you only add the translations before publishing? Could that be it?

msprotz:
oh! perhaps it is, yes -- I haven't refreshed the translations in a long while, and they're not enabled on the -master branch

pianoslum:
Then it should be ok - we just have to be careful that the "between"-translations are included again :-)

PR #1150, opened 2017-02-17
Update footer styling

xi:
I finally good around to work on the footer part of #967. I was actually able to get rid of a lot of code without losing any functionality by just reusing existing parts.

The main change (apart from a lot of styling) is that `.footerActions` is now used in both "normal" messages and quick reply. They also replace the little "open in new window" buttons in the corner of the quick reply textareas. This way there is less code and more UI consitency -- win-win!

I removed one transition (opening a quickreply editor). The issue was that when both textareas are present, the first one still starts out with 100% width and transitions down to 50%. That was just anoying. This issue exists in master.

### DIFF
--- a/content/quickReply.js
+++ b/content/quickReply.js
@@ -95,20 +95,6 @@ function hideQuickReply() {
 
 function registerQuickReplyEventListeners() {
 
-  $('.popout').click(function (event) {
-    let $parent = $(this).parent();
-    let isSelected = $parent.hasClass("selected");
-    let type;
-    if ($parent.hasClass("reply"))
-      type = "reply";
-    else if ($parent.hasClass("replyAll"))
-      type = "replyAll";
-    else
-      Log.assert(false, "There's only two type of textareas");
-    onPopOut(event, type, isSelected);
-    event.stopPropagation();
-  });
-
   $('ul.inputs li.expand').click(function(event) {
     if ($(this).hasClass("selected"))
       return;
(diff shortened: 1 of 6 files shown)


protz:
Thanks for a great piece of work!

I just ran Thunderbird with this for a little while. Here's some feedback.
- there seems to be more gray space below the textarea (when collapsed) than above... here's a screenshot

image

- I actually feel very confused by the removal of the animation... I think it really helped convey that we were either going for reply or reply all
- if there's all four options exposed, the area with the icons feels crammed... any chance we can put them on a 2x2 grid?

image

- the icons feel a little bit bigger than what @elioqoshi suggested in #967
- any chance we can make the icons be on a white background without separation, in case the message is not in "quick reply" mode? actually, this might even work:

image

then you wouldn't have to put the icons on a grid if there's all four of them... what do you think?

xi:
Thanks for the feedback! I think I implemented everything just as you requested in the new batch of commits.

protz:
This looks fantastic, thanks very much.

protz:
Small followup remark: this generates a surprising amount of whitespace

image

I claim that few emails have meaningful content in the bottom right corner. Any way we can make these arrows overlap the bottom right corner of the email, and make them semi transparent, unless when hovered?

Just a thought, let me know what you think!

xi:
The issue with that is that the colors of the message window are user-defined (Preferences > Display > Formatting > Colors…). I, for example, have set the background to black.

I am also not convinced that overlaying the message window is the best idea. Still, I will try.

protz:
Ok, agreed. Maybe I need to give it more time to get used to it...

PR #1153, opened 2017-02-19
RFC: contact initials in UI

xi:
**Please do not merge this yet!**

In #967 there are colors cricles with the initials of the sender on each message. This is a first draft of that.

There are two main challanges here:

1. Get the initials from the name/email
2. Get the color for a contact

Interestingly, I did not find any tutorials of how to do this online. But maybe I just used the wrong searchterms…

About (1): There are many issues here. For one, what we have as a name may well be an email adress, a nickname or some auto generated string (e.g. `name at gmail.com`). Second, the semantics of the name are highly dependent on locale. I did some initial draft algorithm, but we should properly test it before merging.

About (2): With the previous implementation, the same person could have a different color in a different conversation. Since the colors are now much more prominent, I thought it would be better to derive the color from the email address. The resulting colors are ok, but I guess a bit of tweaking would be nice.

I believe that some code is dead after this (e.g. around `freshColor`). Some help with cleaning things up would be appreciated.

### DIFF
--- a/content/stub.xhtml
+++ b/content/stub.xhtml
@@ -802,6 +802,7 @@
       <div class="messageHeader hbox">
         <div class="shrink-box">
           <div class="star"><svg class="icon" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"><use xlink:href="chrome://conversations/skin/material-icons.svg#star"></use></svg></div>
+          <abbr class="contactInitials" style="{{dataContactFrom.colorStyle}}">{{dataContactFrom.initials}}</abbr>
           <span class="author">{{tmpl "contactLabel" dataContactFrom}}</span>
           <span class="to hide-with-details"> {{str "to"}} {{tmpl "contactLabel" dataContactsTo}}</span>
           <span class="recipient-tooltips">{{tmpl "contactTooltip" dataContactFrom}}{{tmpl "contactTooltip" dataContactsTo}}</span>
@@ -1109,7 +1110,7 @@
   <script id="contactLabelTemplate" type="text/x-handlebars-template"><![CDATA[<span
     >{{separator}}</span
     ><span class="tooltipWrapper contact"
-      ><span class="contactName" style="{{colorStyle}}"
+      ><span class="contactName"
         >{{#if star}}&#x2605; {{/if}}{{trim name}}{{#if extra}}
         <label xmlns="http://www.mozilla.org/keymaster/gatekeeper/there.is.only.xul"
           crop="center" class="contactExtra"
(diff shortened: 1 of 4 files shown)


content/stub.xhtml
```diff
@@ -802,6 +802,12 @@
       <div class="messageHeader hbox">
         <div class="shrink-box">
           <div class="star"><svg class="icon" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"><use xlink:href="chrome://conversations/skin/material-icons.svg#star"></use></svg></div>
+          {{#if dataContactFrom.avatarIsDefault}}
+            <abbr class="contactInitials" style="{{dataContactFrom.colorStyle}}">{{dataContactFrom.initials}}</abbr>
+          {{/if}}
```
protz: use an else here?

protz:
Thanks for looking into this! Some comments.
- Deriving the color from the email address is indeed a good idea. I suspect we want different colors for different email addresses, even though two email addresses may belong to the same contact, right?
- Yes, it seems complicated to derive initials and I'm not surprised you haven't found a good reference. https://www.w3.org/International/questions/qa-personal-names provides some context but is not helpful for this particular situation. Your code does what I would've done anyhow -- maybe write a comment along the lines of: "If this is an email address, get the part before the @; then, split across spaces, hyphens and underscores, and capitalize the first letter of the first and last part."

Which things do you need help cleaning up? Happy to give in a hand if needed.

Should I hold off the next release before you merge this in? How do you feel about it? I wanted to use the three-day weekend here to finally push out a new version.

content/stub.xhtml
xi: What's the syntax for that?

xi:
> Which things do you need help cleaning up? Happy to give in a hand if needed.

I think I saw some things that are no longer useful, but I didn't bother at the time. From looking at the code I think `ContactManager._count` could be one of those but there may be more.

> Should I hold off the next release before you merge this in?

I think we should give this some more time. Even if the code is ok, the UX still needs some tuning. I found that the big dark circles draw a lot of attention and may distract from the actual content. Maybe we could post some screenshots and ask for feedback from @elioqoshi.

content/stub.xhtml
msprotz: http://stackoverflow.com/questions/10736907/handlebars-js-else-if I think

protz:
Yes! Screenshots would be very helpful.

Happy to do the cleanups you mentioned after you give me a green light. Also, it looks like some files in `content/i` may no longer be needed.

protz:
I'm getting errors with this patch:
```
Timestamp: 2/20/17, 1:03:27 PM
Error: 2017-02-20 13:03:27	Conversations.Conversation	ERROR	TypeError: initials.toUpperCase is not a function

Source File: resource:///modules/gloda/log4moz.js
Line: 691
```

protz:
More comments:
image
Since there's already "me" next to it, can we stick with JP?

protz:
Oh, perhaps you're getting "me" already as the name, that's why. Happy to patch that up once you're confident the patch is ready.

protz:
That fixes it, thanks

xi:
Ok, I did all the tuning I wanted to do. I guess it is still far from perfect, but I guess it needs someone with more design experience than me.

Feel free to go ahead!

protz:
@elioqoshi want to comment? here's a screenshot

image

protz:
Where are we on this? Honestly is looks good to me and I'd be happy to merge.

xi:
I also used it for some time and did not experience any issues. So feel free to merge!

elioqoshi:
Sorry, just saw this (so many notifications where I don't know where I'm needed :D)

Love this, great job guys! Thanks for this

protz:
Thanks @xi!

PR #1154, opened 2017-02-20
Change star/encryption icon

xi:
This implements the remaining UI changes from #1145 with inline SVG.


protz:
Excellent, thanks. I pushed a followup commit to use `ghead` on OSX in case someone wants to regenerate the icons on a mac machine.

PR #1155, opened 2017-02-21
overlay footer actions on message iframe

xi:
As requested in https://github.com/protz/thunderbird-conversations/pull/1150#issuecomment-280968614 I tried to overlay the footer icons on the message iframe. I am still not sure whether this is a good idea, but with this implementation it should be easier to reason about it.

Here is a screenshot from my setup with a dark message background:

screenshot from 2017-02-21 09 01 15

### DIFF
--- a/skin/conversation.css
+++ b/skin/conversation.css
@@ -655,6 +655,10 @@ span.downwardArrow {
 
 .footerActions {
   float: right;
+  margin-top: -2.3em;
+  background-color: -moz-default-background-color;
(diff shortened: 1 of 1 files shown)


protz:
Thanks -- I've been running with the existing implementation for a while, and I'm not particularly shocked by it anymore. This proposal is likely to cause issues (e.g. hiding relevant information in a given email) so I'm reluctant to go ahead with the proposal; but many thanks for the experiment.

PR #1159, opened 2017-03-05
[WIP] Account for changes in TB beta (50+)

fllp:
It seems the color handling in Thunderbirds mozilla selectors changes from 50.x onwards (i currently use 52.0b4).
`-moz-default-background-color` and `window` yield the same color as the main thunderbird window background (as shown [here](https://github.com/protz/thunderbird-conversations/issues/967#issuecomment-272717709)).  
Although hardcoding the colors should be avoided, there seems to be no alternative from the [color variables available](https://developer.mozilla.org/en-US/docs/Web/CSS/Mozilla_Extensions#%3Ccolor%3E_keywords). Also, by design the message background should always be white, regardless of the native scheme, right?


protz:
I cannot merge this until I issue a release for the current (stable) version of Thunderbird, that is, 45. 

I'm confused about the fix, though. @xi isn't it the purpose of these two colors to be white & black unless the user has a custom OS color theme?

xi:
This is a tough one. I also think that hardcoding colors should be avoided, but thunderbird really makes that hard for us if it changes the keywords without notice.

Is there any thunderbird changelog or bugreport where this change is discussed? Is there anyone working at mozilla who could help us with this?

fllp:
Indeed it is. Shall we move this to an issue instead? Maybe its to soon
for actual code on this.
Skimming a few changelogs didn't yield anything useful so far. It's hard
to find out when the change acutally happened. I think either the
keyword now 'contains' a different color or it has been removed
alltogether, so the next color in the DOM to be inherited is `window`
from the body.
The latter seems to be the case: The DOM inspector indicates a
transparent background when `-moz-default-background-color` is used on
`.message`.
But since i'm quite new to developing with Mozilla apps, i can't say for
sure. Took me some time to find that MDN page of color keywords too.

If there's a mozillian reading this, please enlighten us :)

Am 12.03.2017 um 13:49 schrieb Tobias Bengfort:

elioqoshi:
I'm a Mozillian but that probably is not much helpful as Thunderbird is disattaching itself from Mozilla

fllp:
@elioqoshi: You happen to know where thunderbirds sources are hosted right now? Its hard to find upt to date information in the thunderbird related wikis

fllp:
Even better would be a way to see the complete changelog from 45 onwards. The *What's new* pages/Bugzilla lists unfortunately are not very useful when searching for a specific change

elioqoshi:
@fllp does this help?
https://www.mozilla.org/en-US/thunderbird/45.0/releasenotes/

elioqoshi:
Here are the 52 beta notes:
https://www.mozilla.org/en-US/thunderbird/52.0beta/releasenotes/

msprotz:
This is the list of custom Gecko CSS properties: https://dxr.mozilla.org/comm-central/source/mozilla/layout/style/nsCSSKeywordList.h#38

looks like `-moz-default-background-color` is still in there

xi:
Might be connected to https://bugzilla.mozilla.org/show_bug.cgi?id=1282267. If I understand correctly this changed how `Window` is calculated on windows. @fllp which OS are you using?

fllp:
@protz, @elioqoshi: I already looked at those logs. Unfortunately none
of the changes seem relevant to color handling. And i really can't make
much of the related bugzilla ticket list (its confusing to me).

@xi: I'm running Linux Mint (Ubuntu based) with a GTK DE (Cinnamon). I
think this fix should have landed in the code base for TB 52b as well
since the target release was FX 50. So theoretically it should not be a
problem anymore. But it might be worth exploring once i got some time.

fllp:
I accidentally closed this by deleting the branch. New PR: #1202

PR #1157, opened 2017-03-05
[WIP] Introduce more SVG icons

fllp:
Replace old icons/text links with new icons according to designs from #967.

### polishing needed:

- contact popup
- attachment area

new-icons

PR #1165, opened 2017-03-13
refine icons

### LINKED ISSUE #1164: No icon for preview
image

@xi @fllp any idea what icon may work?

Thanks!

Jonathan

xi:
fixes #1164

@fllp add more icons in #1157. I found some minor issues with that and tried to fix them:

- Some icons were inside of inline `<a>` elements, leading to extra whitespace. I fixed that by setting `display: inline-block` on these links.
- The "more" dropdown was not at the same position as the button.
- The "details" and "more" icons were rather small


xi:
By the way, I start to fell like we might overdo it with the icons. Adding icons is fine, but replacing text by icons might actually reduce usability, especially if the icons are no perfect matches. The contact popup might be a place where the icons alone may not be sufficient.

elioqoshi:
I agree that text is sometimes the best "icon" So let's not overdo it. However, if space permits, I'd support Icon accompanied by text.

fllp:
@xi: good stuff! #1157 was still very rough in some areas.

As for the text vs. icons: Personally I'm allright with relying on
tooltips for icons whose function are not immediately clear to me. But
from a UX point of view I agree  with you that icons should be as clear
as possible and have text next to it when necessary.

But to be honest: The old underlined text links were not so beautiful...

content/stub.xhtml
```diff
@@ -837,7 +837,7 @@
             <span title="{{fullDate}}">{{date}}</span>
           </span>
           <span class="replyLinkWrapper">
-            <a href="javascript:" class="link replyMainActionLink" style="display:none"></a>
+            | <a href="javascript:" class="link replyMainActionLink"></a>
```
pianoslum: I don't know much about xhtml, but does that '| ' at the beginning belong there?

xi: It had been there before and was removed in https://github.com/protz/thunderbird-conversations/pull/1157/commits/63a6828f163cc97e6ac3e762db18fd1f7650639a. Apart from that, yes, I think this is a simple text node and totally valid.

msprotz:
Sorry for the delay, this is great. Thanks!

PR #1168, opened 2017-03-14
include certificate icon from font-awesome

xi:
@fllp suggested to use this icon in https://github.com/protz/thunderbird-conversations/issues/1164#issuecomment-286467342. It is not ideal. Font awesome does not officially provide an SVG version, but there is a [separate project](https://github.com/encharm/Font-Awesome-SVG-PNG) which does that. Also note that the view-box is different for material and awesome icons.

I am not completely convinced, so it is really up to you to merge it or not.


fllp:
@elioqoshi: Is there some sort of SVG version somewhere for the 'new' icon you used in your mockups?

fllp:
Here's another thought: `new_releases` new releases looks much like the sticker in the design drafts, except for the exclamation mark. We could remove that from the original SVG and add it to the icon collection SVG under a different name. It would be the same view box and wie would not have to rely on external sources

xi:
> Here's another thought: new_releases

That's a good one! I think we could even use new_releases without modification.

I thought it was silly to open a new pull request for each new option. So I pushed both an unmodified and modified version of new_releases to this branch. You can easily change icons by modifying stub.xml (see [c0b7421](https://github.com/protz/thunderbird-conversations/pull/1168/commits/c0b742161d16e50bba780a7dc1cdc2b42f578b6e#diff-47cf72757acbc915d29a40ad9ffa10b7L1439)).

If I had to decide, i would probably just use an unmodified new_releases.

msprotz:
Hey,

Can you post screenshots and refresh the pull request? That'll help me merge.

Thanks for the pull request!

Jonathan

fllp:
@protz, here you go:

### Fontawesome icon
(currently used in stub.xhtml)

unread fontawesome

read fontawesome


### Material Icon
(unmodified `new_releases` icon)

unread vanilla material

read vanilla material

### Modified material icon

unread modified

read modified

msprotz:
The `!` is confusing and reminds me of the "important" status that appears in many mail clients. I'd be in favor of either the fontawesome icon or the modified material icon.

fllp:
I agree. The modified `new_releases` icon looks best. Plus, there should
be no hassle with differnent licenses, right?

protz:
I don't think so as long as we make it clear in LICENSE that we're 
re-using parts from other projects.

xi:
replaced by #1174

PR #1173, opened 2017-03-27
Reapply attachment icon rotation

fllp:
This got reverted somewhere along #1165 or #1166

### DIFF
--- a/skin/conversation.css
+++ b/skin/conversation.css
@@ -623,6 +623,10 @@ a img {
   color: #666;
 }
 
+.attachmentIcon .icon {
+  transform: rotate(-50deg);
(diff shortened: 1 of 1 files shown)


xi:
I removed it intentionally. I should probably have asked. What is this for? For me the rotated icons look extremly strange. (I can add a screenshot later)

fllp:
Oh i see. Well, according to the mockups it should be rotated. But in
the end, its @protz's decision to make.

xi:
I realized now that this is about the attachment icon in the header. I was confused because I thought it was about the preview/save/open icons in the footer. So nevermind.

Issue #1172, opened 2017-03-27
[Redesign]: Deviations from mockup


fllp:
This may be nit-picking but i noticed some differences in the current implentations of the redesign. Mostly the deviations were introduced with #1165. How should they be treated? Personally, i'd prefer the mockup versions.  
Comparisions below. @protz, @xi and @elioquoshi: What's your opinion

### Duplicate reply all action

- applies to reply to list as well
- should at least be an icon instead of text link   
- IMHO the `|` as separator doesn't fit into the new design

Original mockup:  
![mockup-reply-all](https://cloud.githubusercontent.com/assets/15898292/24366084/52a7f632-1318-11e7-83cc-e113d5aedf74.png)

Current state:  
![reply-all](https://cloud.githubusercontent.com/assets/15898292/24365987/0a1d6eec-1318-11e7-9d6f-5ba54c078f55.png)

### Attachment icons

I just dont like the floppy icon :stuck_out_tongue:

Original mockup:
![mockup-attachments](https://cloud.githubusercontent.com/assets/15898292/24366477/862c4d72-1319-11e7-8080-efd7bbd695ea.png)

Current state:
![attachments](https://cloud.githubusercontent.com/assets/15898292/24366464/7e862d54-1319-11e7-95f1-30db4a494168.png)

protz:
Ok to replace floppy with downarrow.

The reply all link at the top is important, because sometimes emails are
super long and you don't want to scroll all the way to the bottom. Can you
make it an icon?

Thanks!

Jonathan

protz:
Also, the (i) info icon seems a little too big now...? I'm wondering if
that's something we can fix

protz:
Can we also replace the "preview" icon with an eye? I'd think this is the standard idiom for preview icons.

xi:
@fllp thanks for the summary!

> The reply all link at the top is important, because sometimes emails are super long and you don't want to scroll all the way to the bottom. Can you make it an icon?

+1

> Ok to replace floppy with downarrow.

For me the downarrow is confusing. I guess it is carried over from webservices like gmail, where you need to download the attachment. But in thunderbird, the download metaphor does not make a lot of sense.

The floppy on the other hand fitting and on top of the much more specific than an arrow.

> Can we also replace the "preview" icon with an eye? I'd think this is the standard idiom for preview icons.

See https://github.com/protz/thunderbird-conversations/issues/1164#issuecomment-286216396. (I think I closed that issue by accident btw)

[...]

PR #1175, opened 2017-03-28
replace mainActionLink with icon

xi:
see #1172

### DIFF
--- a/content/stub.xhtml
+++ b/content/stub.xhtml
@@ -837,7 +837,7 @@
             <span title="{{fullDate}}">{{date}}</span>
           </span>
           <span class="replyLinkWrapper">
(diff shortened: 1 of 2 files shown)


protz:
Good! However, I notice that the icon itself (e.g. pdf) is now no longer clickable to preview the attachment... was this intentional?

xi:
> I notice that the icon itself (e.g. pdf) is now no longer clickable to preview the attachment... was this intentional?

I cannot reproduce that issue.

protz:
I just tried: at revision 395a409 (before I merged your branches), I 
could click on the PDF icon of an attachment (the file icon) and this 
would preview the PDF. Now, the icon is no longer clickable.

PR #1176, opened 2017-03-29
make sure that action buttons are centered

### LINKED ISSUE #987: Conversation buttons icons not centered
I'm wondering why no one has reported this yet, but on my MacBook Pro running 10.10.4 and Thunderbird 38.1.0 with TbConversations 2.9.1 the conversation buttons are off, see screenshot.
screen shot 2015-07-16 at 11 50 27

xi:
I can not test on a mac either, but this should fix #987.

### DIFF
--- a/skin/conversation.css
+++ b/skin/conversation.css
@@ -147,7 +147,8 @@ a img {
 }
 
 .actions button {
-  width: 30px;
+  min-width: 30px;
+  text-align: center;
   height: 24px;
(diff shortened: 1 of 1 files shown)


protz:
Is this relevant anymore? We no longer have borders on icons...?

xi:
There are borders on hover. I think the change makes sense either way and it closes an issue, so why not merge it?

PR #1189, opened 2017-05-15
Fixes #1184: Change process functions for gloda to generators to account for changes made in Thunderbird bug 1292569.

### LINKED ISSUE #1184: Indexing messages broken after Thunderbird 52 update
Hello,

After Thunderbird 52 update, Conversations add-on started acting weird as described in some of the previous issues: Sent emails do not show in the threads, Emails that should not be on the thread appearing etc.
It turned out that the Message Index Database is corrupted and after following steps on rebuilding database from [here](https://support.mozilla.org/t5/Troubleshooting/Rebuilding-the-Global-Database/ta-p/15801), Indexing Messages was stuck and didn't want to complete.
After disabling Conversations, Indexing Messages went through just fine and on re-enable, Conversations worked as intended, until new email is received and Indexing Messages process was stuck again so the process described above had to be repeated.
As a temporary solution, I have downgraded to Thunderbird version 45.8 [here](http://releases.mozilla.org/pub/thunderbird/releases/45.8.0/win32/en-US/) and disabled automatic updates as instructed [here](http://www.cnx-software.com/2011/10/05/disable-auto-update-for-thunderbird-and-firefox/).

Please let me know if any additional info is needed.
Thanks,
Pavle

Standard8:
Bug 1292569 made a change to gloda's indexer.js that means it now requires proper generators to work correctly.

I noticed this due to a `TypeError: this.activeIndexer.return is not a function` message on the error console.

This change gets the indexer running again in 52.

### DIFF
--- a/modules/plugins/glodaAttrProviders.js
+++ b/modules/plugins/glodaAttrProviders.js
@@ -89,7 +89,7 @@ let AlternativeSender = {
     });
   },
 
-  process: function _AlternativeSender_process (aGlodaMessage, aRawReps, aIsNew, aCallbackHandle) {
+  process: function* _AlternativeSender_process (aGlodaMessage, aRawReps, aIsNew, aCallbackHandle) {
     try {
       let alternativeSender = PluginHelpers.alternativeSender(aRawReps);
       if (alternativeSender)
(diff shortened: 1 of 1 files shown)


rajeshstarlite:
I tried fix but it didn't work for me.

Gregory-K:
tested and working ... so far (Win10, MacOS)

yalesov:
correct fix, works perfectly, thanks!

chiwanpark:
Works perfectly. Thanks for the valuable patch.

PR #1190, opened 2017-05-15
Update the querySelector for looking at bugzilla emails, so that selecting them works again.

Standard8:
This seems to be an issue on latest master - selecting a Bugzilla message results in an email view, that is mostly complete, but significantly the email doesn't get marked as read.

Looks like something changed the class names and broke the selector rules, and in turn this broke updating the display.

This seems to fix it for me.


protz:
Yes, this is good.

PR #1194, opened 2017-05-26
Fix an issue displaying initials that contain utf-8 characters that causes the message display to hang.

Standard8:
One person I get emails from has changed their name field to:

first last 🐧

Currently, this causes conversations to hang and throw out a bunch of errors. This is because the initials handling code is not taking account of the utf-8 character.

developer.mozilla.org had a suggest way around this, so I've added that to the code.


protz:
Thanks Mark. The past couple months have been a little hectic, but this is afternoon is the first quiet Sunday in a while, so I'm looking at all your pull requests right now. Cheers!

PR #1198, opened 2017-06-13
readme: replace reference to the old name

SimJoSt:
finishing my work from #1126


Standard8:
Looks great, thanks for correcting this.

PR #1199, opened 2017-06-15
Enable ESLint using the mozilla/recommended rules.

Standard8:
@protz, this does the minimum necessary to enable ESLint on the repository.

I've included eslint-plugin-mozilla as I think we'll want that, and I added the mozilla/recommended configuration as a default one. I've turned off the rules where we fail.

The code changes are mainly things where we were using array comprehensions & other non-standard javascript items. The Firefox code base has moved away from these, so I think it is sensible to do that here as well.

Can you enable Travis CI on this repository? Just visit https://travis-ci.org/ and sign-in, then go to account and flip the switch for the repository.

Once we've got one test run reported, it would be nice to turn on master as a protected branch in github and require the tests to pass to merge PRs.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,10 @@
+# Don't ignore .eslintrc.js
+!.eslintrc.js
+
+# Third party libraries.
+content/js/**
+content/gallery/js/jquery.lightbox-0.5.js
+
+# Ignore sub-module directories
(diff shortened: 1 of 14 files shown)


protz:
Thanks! That'll be very helpful in the long run. I've done the required changes on the Travis side.

PR #1203, opened 2017-06-20
Attachment icons: Replace floppy with down arrow

fllp:
Partial fix for #1172


fllp:
@protz or @standard8: can you take a look at this?

fllp:
Thank you :)

PR #1204, opened 2017-06-20
Option icons: Decrease scaling factor

fllp:
Scales down icons for mail header toggle and more menu by 10%  
This should compromise between too small icons and too much 'zoom'
Partial fix for #1172


fllp:
@protz, @xi, @standard8: is this a good middle ground for you?

xi:
Works for me.

PR #1223, opened 2017-08-28
Fix #1220: Update stdlib submodule to incorporate fixes for Thunderbird 57 compatibility.

### LINKED ISSUE #1220: Handle interfaces that have been removed from Thunderbird Daily (57)
As per tb-planning, there's some interfaces that have gone away in 57 that we'll need to fix.

As we support multiple versions, we should generally do a check to see if the old interface is available, and use that, otherwise switch to the new one.

Interfaces I know of so far:

- [ ] nsIScriptableDateFormat (in stdlib) (replace with https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/DateTimeFormat)
- [ ] nsILocalFile (in pdfjs)

xref:

- https://bugzilla.mozilla.org/show_bug.cgi?id=1347507

Standard8:
This pulls in the actual change we need from stdlib.


Standard8:
I'm not going to land this yet, until I've resolved an issue with backwards compat for TB 52.

Standard8:
I forgot about this PR, I landed this in eaba9f15d7a4d0d5445bb298c97a986b7d85b02c

PR #1235, opened 2017-09-13
Archive/delete pref label improved, fixes #1225

### LINKED ISSUE #1225: Change title in pref for "Toolbar buttons"
I was searching forever for how to make delete actually delete both received and sent emails in a conversation (and I'm frustrated about years of undeleted sent emails up to now).

The preferences say "Toolbar buttons" as the title, and only reading the fine text makes it clear this is not about having buttons or not, this is about their behavior… some better title would be good. At least something like "Toolbar button behavior" or "Archive/delete behavior"

wolftune:


protz:
Thanks!

PR #1244, opened 2017-10-01
Fix(general): Enable the no-undef rule for most of the add-on.

Standard8:
This enables the no-undef rule for everything apart from content/. I think it may be possible to clean up some of the imports further, but we'll see to that later.

Additionally note that as eslint-plugin-mozilla doesn't support Thunderbird's modules, I'm changing the style of import to be explicit variables.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,5 +1,6 @@
 "use strict";
 
+/* eslint-env node */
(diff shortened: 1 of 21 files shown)


Standard8:
Merging to get this landed as it'll help protect against future issues.

PR #1245, opened 2017-10-01
Clean up some deprecated APIs usage

Standard8:
String.x is deprecated: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Deprecated_String_generics

Also folder.prettiestName is reporting as deprecated as well.

There's still others to fix, but we can leave those for later.

@protz please can you look over this, especially the String.* changes.


Standard8:
protz hasn't been about, and I'm fairly confident these are correct, so going with landing.

protz:
Thanks Mark!

PR #1254, opened 2017-11-17
allow resizing of textareas

vhdirk:
I added a resize handle to quick reply since it bothered me that the textarea was always way too small on my screen


vhdirk:
It's true that the grippie image itself is not showing. I don't know why that is. I'll have to check. However, the textarea border is a bit wider and you can still grab and resize with that.
I was wondering why the standard browser grippie was not showing, but I did not know you could enable/disable that from css.

The patch only allows vertical resizing.

You know, the sole reason I made this is that 'enlarge-as-you-type' does not seem to work on my system. And it bugged me that I had this huge blank space below the quick reply window that was never used...

Standard8:
Closing this out, as there wasn't follow up, but also with the current UI changes I think this might be simpler once we get quick reply running again. I'll try and bear it in mind when we add that back.

Issue #1116, opened 2016-09-13
Switch from BzAPI to the Bugzilla native REST API


edmorley:
Since the former is deprecated and 2-10x slower in many cases. See:
https://bugzilla.mozilla.org/show_bug.cgi?id=1301541
https://bugzilla.mozilla.org/show_bug.cgi?id=1035804

https://github.com/protz/thunderbird-conversations/blob/b989e21b7b0e8340286d8bde81e9a3d3cb508798/content/stub.compose-ui-bz.js#L49-L54

protz:
Hi Ed,

Sorry for the delay in responding. I have close to zero time for the addon except for the bug triaging and occasional critical bugfix. Do you think you could submit a pull request? I'd love to integrate that change...

Thanks,

Jonathan

Standard8:
We're going to be dropping the bugzilla integration, hence closing out.

PR #1272, opened 2018-04-10
Add BzAPI definition for bugs.kde.org

### LINKED ISSUE #1271: BzAPI for bugs.kde.org
Hi,

The notice I got in Thunderbird led me to the [wiki](https://github.com/protz/thunderbird-conversations/wiki/BzConversations) page and I can confirm that the API endpoint of bugs.kde.org is just `https://bugs.kde.org/rest`.

For instance, I could run
```
curl https://bugs.kde.org/rest/bug/300000
```
...to fetch the details of this bug. And they also allow generating API keys at https://bugs.kde.org/userprefs.cgi?tab=apikey

Frederick888:
Closes #1271

### DIFF
--- a/content/stub.compose-ui-bz.js
+++ b/content/stub.compose-ui-bz.js
@@ -47,12 +47,18 @@ if (!("cookies" in Services)) {
 }
 
 let gBugzillaAPIs = {
-  "https://bugzilla.mozilla.org/":
-    "https://api-dev.bugzilla.mozilla.org/latest/",
-  "https://landfill.bugzilla.org/bzapi_sandbox/":
-    "https://api-dev.bugzilla.mozilla.org/test/latest/",
-  "http://bugs.kde.org/":
-    "https://bugs.kde.org/rest/",
+  "https://bugzilla.mozilla.org/": {
+    url: "https://api-dev.bugzilla.mozilla.org/latest/",
+    type: "bzapi",
+  },
+  "https://landfill.bugzilla.org/bzapi_sandbox/": {
+    url: "https://api-dev.bugzilla.mozilla.org/test/latest/",
+    type: "bzapi",
+  },
+  "http://bugs.kde.org/": {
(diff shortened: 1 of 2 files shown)


msprotz:
Thanks for the pull request! Have you confirmed that this works for you?

Frederick888:
@msprotz Well, it turns out that they are now using the new [native REST APIs](https://wiki.mozilla.org/Bugzilla:REST_API) instead of BzAPI so it didn't work out of the box. I tried the BzAPI [compatibility layer](https://wiki.mozilla.org/Bugzilla:BzAPI:CompatLayer) but it seems that they've dropped this functionality as well.

BUT, the old and new APIs are actually rather similar and the new API still supports cookie-based authentication. So I have just pushed another commit to make the extension compatible with the new API (a bit hacky tho). I have manually tested it by adding a new comment to an issue.

Is commenting the only integration we have for Bugzilla?

msprotz:
Can we upgrade everyone to use the new REST APIs? I think the BzAPI is going to be retired anyhow, and I would suspect Mozilla's Bugzilla also has a REST endpoint. See #1116  -- that way, your patch would be even simpler and could retire the old code :)

Frederick888:
@msprotz To me it's like why not. And it would allow us to provide general Bugzilla support as well since the API URL will be consistently `http[s]://bugzilla.example.com/rest` (of course an exception table can be maintained).

If you could provide some mentoring I guess I can take the task.

protz:
Hello @Frederick888  -- got back from vacations and catching up.

My suggestion would be to just take your patch, and remove the "type" field, and make everything go through the newer REST API. In other words: don't if-then-else on whether it's a bzapi or REST endpoint, and assume that it is always REST.

What do you think? This should be a minor change to your patch, and I believe would be a good way to bring Conversations to use the newer APIs.

Standard8:
This never seemed to get completed and we're going to be dropping the bugzilla integration soon.

PR #1286, opened 2018-07-13
Update Outlook blockquote detection for newer versions

Nephyrin:
Based on my painful inbox, newer outlook versions slightly changed the shade of this horizontal line, so we need to refine the check a little.

The HTML still looks entirely generic with nothing better to select against, though.

### DIFF
--- a/modules/quoting.js
+++ b/modules/quoting.js
@@ -135,7 +135,8 @@ function convertOutlookQuotingToBlockquote(aWin, aDoc) {
   trySel(aDoc, ".OutlookMessageHeader");
   for (let div of aDoc.getElementsByTagName("div")) {
     let style = aWin.getComputedStyle(div, null);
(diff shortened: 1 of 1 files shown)


Nephyrin:
And, for the record, here's what I'm seeing in the message source at the break between the message and the quoted text:

```html
<p class="MsoNormal"><span style="color:#1F497D">Blah blah last line of message<o:p></o:p></span></p>
<p class="MsoNormal"><span style="color:#1F497D"><o:p>&nbsp;</o:p></span></p>
<p class="MsoNormal"><span style="color:#1F497D"><o:p>&nbsp;</o:p></span></p>
<div>
<div style="border:none;border-top:solid #E1E1E1 1.0pt;padding:3.0pt 0in 0in 0in">
<p class="MsoNormal"><b>From:</b> Sender Name <br>
```

PR #1291, opened 2018-08-08
enigmail 2.0 support

shirosaki:
Fix for Thunderbird 60 and Enigmail 2.0.8.

- verify signed PGP MIME messages
- decrypt verify inline PGP messages

Quick reply with PGP doesn't work.

### DIFF
--- a/modules/plugins/enigmail.js
+++ b/modules/plugins/enigmail.js
@@ -108,7 +108,8 @@ let hasEnigmail;
 try {
   /* globals EnigmailCommon, EnigmailCore, EnigmailData, EnigmailLocale,
              EnigmailFuncs, Enigmail, EnigmailPrefs, EnigmailDialog,
-             EnigmailConstants, EnigmailRules */
+             EnigmailConstants, EnigmailRules, EnigmailArmor,
+             EnigmailDecryption */
   ChromeUtils.import("resource://enigmail/core.jsm");
   ChromeUtils.import("resource://enigmail/data.jsm");
   ChromeUtils.import("resource://enigmail/dialog.jsm");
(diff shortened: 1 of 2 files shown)


Standard8:
Hi @shirosaki I've just merged a patch from the author of enigmail that fixes at least decrypting email, can you see if it fixes these other issues as well, or if we do need to do the more complex work here?

shirosaki:
I've rebased on master.
I've fixed quick reply with PGP and inline PGP decryption bug.

shirosaki:
@Standard8 
I'm using this patch and it works fine for me.
Could you review this?

modules/plugins/enigmail.js
```diff
@@ -173,6 +179,16 @@ if (hasEnigmail) {
     global.EnigmailConstants = EnigmailCommon;
   }
 
+  let isEnigmail2 = false;
```
Standard8: I think we should drop pre-enigmail 2 support, but I'm happy not to do that in this PR, we can do a follow-up. I've filed #1305 to remind us.

modules/quoting.js
```diff
@@ -146,8 +146,8 @@ function convertForwardedToBlockquote(aDoc) {
       let txt = child.textContent;
       let m = txt.match(re);
       if (child.nodeType == child.TEXT_NODE
-          && txt.includes("-----BEGIN PGP")
-          && txt.includes("----END PGP")
+          && !txt.includes("-----BEGIN PGP")
```
Standard8: What is this change for?

shirosaki: https://github.com/protz/thunderbird-conversations/commit/100641dbd663a3cebab55e279cc47e4078565821#diff-e54960bec9e47f840349eca45e3dcd94L171
Your commit changed that. It breaks inline pgp decription.
Not including is right.

Standard8: Thank you for spotting that. Looks like I must have changed it manually and got it wrong :-(

I'll try and find time to look at the patch one more time in the next day or so.

modules/plugins/enigmail.js
shirosaki: I agree.
There are a lot of code for old Enigmail.
I'll remove old code at the next time.

Standard8:
Thanks for the patch. Sorry for the delay in getting to review this.

PR #1293, opened 2018-08-11
Adopt decryption to API of Enigmail 2.0

pbrunschwig:
The API for Engimail.msg.messageDecryptCb has changed with Enigmail v2.0. This patch adopts to the change (and fixes the Enigmail bug https://sourceforge.net/p/enigmail/bugs/894/).


Standard8:
Thanks for the patch @pbrunschwig. Do you know if this fixes quick reply (xref #1291)?

Issue #1308, opened 2018-09-16
Support Thunderbird builds post TB 60


CyberMan571:
* **Thunderbird version:**
Name 	Thunderbird
Version 	64.0a1
Build ID 	20180915120746
* **Thunderbird Conversation version:**
Thunderbird Conversations	2.14.4
* **Operating System:**
OS 	Windows_NT 10.0

The Thunderbird Conversation version of 2.14.4 does not work ing the thunderbird daily build

the progress of the extension in first setup is not continue

like it below

![image](https://user-images.githubusercontent.com/30759335/45592084-bccda380-b96c-11e8-9ad9-81372039259f.png)

Standard8:
This is quite possible, the core of Thunderbird is changing a lot at the moment, and it is highly likely that things will get broken.

When I do get a bit of time, my work is more likely to be for improving the architecture. So if anyone wants to contribute patches, that would be most welcome.

Preferably, patches would support TB 60 as well as the next versions, but there may come a time where we have to break that. 

https://wiki.mozilla.org/Thunderbird/Add-ons_Guide_63 is sometimes a useful resource to reference, though it isn't always updated straight away.

scheinercc:
Just to clarify is this a tracking bug for Daily _and Beta_ now? Thanks

Standard8:
Yes, although #1323 made some improvements for 66, that have landed on master, there's a few regressions that need addressing before we can release it.

The next major version of Thunderbird is not until 68 (which will be around July), so until then, although I'd like to support latest beta if possible, it isn't a priority. Some of it may also be complex, since supporting both 60 + latest may also get complicated from a compatible with both point of view.

As I said before, patches are welcome.

Standard8:
Just to note, the changes that have landed have caused #1346 so we can't release them at the moment until we fix that.

Standard8:
I've started getting master updated to work with 68. The basics work, but quick reply, contact icons and probably a few other things are broken at the moment.

Some of the work to switch to react may benefit the issues around #1346, but switching is quite a long task, so we may need to figure out a workaround to get our images to display in the meantime.

Standard8:
2.15.2 has just been released & approved which now works with Thunderbird 68 betas (nothing earlier).

This has featured some significant rework that is still in progress but helps to fix a number of issues.

[...]

Issue #1255, opened 2017-11-19
Conversations doesn't work with latest Thunderbird (60)


Standard8:
* **Thunderbird version:** 59.0a1
* **Thunderbird Conversation version:** 2.13.4pre
* **Operating System:** Mac

**What did you do?** Tested in latest daily

**What did you expect to happen?** Conversations to work

**What actually happened?** It didn't work, errors on the console:

NS_NOINTERFACE: Component returned failure code: 0x80004002 (NS_NOINTERFACE) [nsIPrefBranch.getComplexValue]

koolx:
Conversations dont evenm work in the latest standard TB 52.5.

Standard8:
@koolx Please can you file that as a separate issue, and describe what exactly you are seeing. Commenting on an existing unrelated issue is not helpful.

koolx:
How is it an unrelated when my issue is the same as yours that TC dont work on the current version of TB?

Standard8:
@koolx You mentioned Thunderbird 52, I mentioned Thunderbird 59 - they are very different and 59 is newer and has had more development than 52 (and 59 is a pre-release version) so it isn't surprising that TC might not work with the latest pre-release version.

Also, my first comment is listing a specific message that is occurring in the console (Tools -> Developer Toolbox -> Error Console), I very much doubt you are seeing that, or have even checked that (in this case, I know it is only going to happen in TB 58 & later as that's where the interface was removed).

I also can expect that TC is working fine for most of its users in 52, as we haven't been getting reports of issues. Therefore I suspect this is something else, and hence we need full details in a separate issue please.

koolx:
ok

rnc:
@koolx @Standard8  I run 52.4.0 on Fedora 27 without issues.I downloaded a 52.5.0 and it starts up fine with Conversations functioning as normal.

koolx:
ok

koolx:
e

koolx:
@rnc Heres the problem:

I got 2 accounts in TB. When I send email from account A to B I get a notification in B's inbox that I got it. I then reply from account B and I get a notification in inbox A that I got it. But when I reply to account B, I dont get a reply NOR a notification in B's inbox. Whats even stranger is that the email and notification end up in account A's inbox, not B's. 

This error goes away when I disable Conversations.

This is a bug. Please try it on your end if you got more than one account set up. Conversely, I can generate and send you an error log, if possible. 

Let me know.

[...]

PR #1315, opened 2018-10-11
Fix for Thunderbird 64

shirosaki:
I have fixed some issues for Thunderbird 64.
Messages are visible in conversations for now.
Many features are still broken.

- innerHTML does not work. Use DOMParser.
- XPCOMUtils.generateQI is deprecated. Use ChromeUtils.
- chrome url for images causes security errors. Use css.


content/stub.compose-ui.js
```diff
@@ -817,7 +816,7 @@ function nsAttachmentOpener() {
 
 nsAttachmentOpener.prototype = {
 
-  QueryInterface: XPCOMUtils.generateQI([
+  QueryInterface: ChromeUtils.generateQI([
```
Standard8: Do you think we could make this work for the current TB 60 series & 64 and later as well?

I think at some point we'll need to break support lines, but for now it might be useful to keep them.

content/stub.xhtml
```diff
@@ -18,11 +18,6 @@
     href="chrome://conversations/skin/tokeninput.css" />
   <link rel="stylesheet" type="text/css" media="print"
     href="chrome://conversations/skin/print.css" />
-  <script type="text/javascript">
-    <![CDATA[
-      document.allowUnsafeHTML = true;
```
Standard8: xref https://github.com/protz/thunderbird-conversations/issues/1255#issuecomment-379575588

Also https://bugzilla.mozilla.org/show_bug.cgi?id=1450315#c8

Basically, we need to move away from jQuery now. I've just filed #1317 to track that.

modules/contact.js
```diff
@@ -39,6 +39,7 @@
 var EXPORTED_SYMBOLS = ["ContactManager", "Contacts", "defaultPhotoURI"];
 
 ChromeUtils.import("resource:///modules/StringBundle.js"); // for StringBundle
+ChromeUtils.import("resource:///modules/DisplayNameUtils.jsm"); // for DisplayNameUtils.getCardForEmail
```
Standard8: nit: Please use the alternate format here, as it is better for ESLint to detect the globals:
```suggestion
const {DisplayNameUtils} = ChromeUtils.import("resource:///modules/DisplayNameUtils.jsm");
```

Standard8:
General note, this is helping towards #1308

content/stub.compose-ui.js
shirosaki: I found that Enigmail uses some code for compatibility.
We could make something like this.
https://gitlab.com/enigmail/enigmail/blob/master/package/tb60compat.jsm

Standard8: Ooh, thanks for pointing that out. I'd be happy with something like that. Nice and clear as to where we're adding compatibility fixes as well.

Standard8:
@shirosaki I'm sorry, for not seeing the updates here... github doesn't seem to notify if there's no comments :-(

Do you want to include the compatibility fix for TB 60, or should I look into that?

shirosaki:
@Standard8 I cannot find a clean way to include the compatibility fix for TB 60. Please look into it.
- XPCOMUtils.importRelative
- XPCOMUtils.generateQI

shirosaki:
Some fixes are needed for Enigmail 2.1. A lot of refactoring is done in Enigmail.

modules/plugins/enigmail.js
```diff
@@ -106,29 +106,23 @@ let Log = setupLogging("Conversations.Modules.Enigmail");
...
-  try {
-    ChromeUtils.import("resource://enigmail/enigmailCommon.jsm");
-    ChromeUtils.import("resource://enigmail/commonFuncs.jsm");
-    hasEnigmail = true;
-    Log.debug("Enigmail plugin for Thunderbird Conversations loaded!");
-  } catch (e) {
-    hasEnigmail = false;
-    Log.debug("Enigmail doesn't seem to be installed...");
-  }
+  Log.debug("Enigmail doesn't seem to be installed...");
```
Standard8: I'd suggest adding this here, so that we don't get extra warnings on TB 60 builds:
```suggestion
  Log.debug("Enigmail doesn't seem to be installed...");
  var EnigmailConstants = {};
```

shirosaki:
Thanks for the patch.
That looks fine.
I think that my Enigmail patch doesn't work with TB60. Fix for TB66.

modules/plugins/enigmail.js
shirosaki: Fixed differently.
Move `nsIEnigmail = EnigmailConstants;` into `if (hasEnigmail) {`.

content/stub.compose-ui.js
shirosaki: Use compatibility code is in misc.js of https://github.com/protz/thunderbird-stdlib.

shirosaki:
Rebased and pushed. Enigmail fix is a separated commit.

content/stub.xhtml
shirosaki: Dropped jQuery features related to `innerHTML`, but this still dependents on jQuery.

Issue #1324, opened 2018-11-15
Inline reply does not allow for attachments


parrenin:
* **Thunderbird version:**  60.2.1
* **Thunderbird Conversation version:** 2.14.5 
* **Operating System:** Debian Stretch

**What did you do?**

I replied inline to an email ("Reply" button at the bottom of the message) and then I clicked on "add attachment". 

**What did you expect to happen?**

A file chooser window should appear to select the file to attach.

**What actually happened?**

Nothing.

Standard8:
Fixed by #1329.

I've just posted a new version of the add-on. It should be published soon (once reviewed).

Standard8:
Thanks to @dethmetaljeff for fixing this.

PR #1329, opened 2018-12-05
replace filePicker.show with open since it's been deprecated

dethmetaljeff:
Fix for #1324


Standard8:
Thank you for the changes. They work nicely. I added a minor change to fix to 2-space indentation as that's what we're generally using now.

Sorry for not responding earlier, I've been busy with travelling.

Issue #1303, opened 2018-08-28
Need to hit enter key twice to add a line feed in Quick Reply.


sahil48:
* **Thunderbird version:** 60.0
* **Thunderbird Conversation version:** 2.14.4
* **Operating System:** Windows 10

**What did you do?**
I hit the enter/return key twice using the Quick Reply Feature.

**What did you expect to happen?**
I expected to see two line feeds
**What actually happened?**
I only see one line feed and have to press enter a third time to add the second line feed.

Standard8:
Yes, I see this as well.

touch444719:
Any news regarding this issue ? 

The "quick reply" feature is the only feature that still block me to update to V60.
This issue as well as when you "clic" the quick reply field, it doesn't automatically put your cursor right on top to directly write your reply (with some spaces already fill in between your cursor and the last reply).

Basically exactly what I have right now with the 52.9.1 thunderbird and the 2.14.4 conversation addon.

Thank you for your work so far tho !

ZizzyZizzy:
Ditto here. You have to hit enter twice for a CR+LF, and three times if you want a new paragraph.

This is new behavior with the new version from Jan. Very annoying.

koolx:
Exactly. Why the hell would the dev make the addon behave this way?? 
I've never seen a program where you hit enter twice to produce a new 
sentence and 3 times to start a new paragraph. Could you correct this, 
its very damn annoying and causes many inexcusable typing errors.

koolx:
Another issue: Please convert military time to standard time.

Standard8:
Unfortunately "the dev" doesn't have much time to work on conversations, and this is the occasionally bit of time in the evenings after having done a day's work, decorating, looking after the kids etc. At the moment, "the dev" is working on a massive overhaul so there's at least a chance of it continuing to have basic functions for when Thunderbird 68 is released (soon).

I expect fixing this to be a multi-hour debugging session as the reason this is happening is not obvious - I expect it to be in the core of Thunderbird somewhere, in a component I'm not familiar with. I will try and fix it at some stage, but knowing it likely really needs a single deep session is something I can't allocate time for at the moment.

I would like to remind you that this is free and open source software. You are quite welcome to try and help fix it or find other people that might. If you continue with this attitude, I will be reporting you.

koolx:
> 
> 
[...]

PR #1331, opened 2018-12-18
use designMode instead of WindowEditable

dethmetaljeff:
Fix for #1303.  Looks like you were using designMode before but switched to makeWindowEditable...not sure why that was done.  Everything seems to work fine for me using designMode and it fixes the bug...so, maybe it's right?


Standard8:
Closing out as it seemed to be the wrong approach and no follow-up.

With what I've been able to see in 68 this may have gone away, but not totally sure. Will investigate when I work on re-enabling it.

Issue #1317, opened 2018-10-20
Remove usage of jQuery


Standard8:
Per previous discussions and especially a couple of [comments](https://bugzilla.mozilla.org/show_bug.cgi?id=1450315#c8) on [bugs](https://bugzilla.mozilla.org/show_bug.cgi?id=1433871#c14), we should move away from using jQuery in conversations.

There's probably several parts to this:

- [x] The gallery view (content/gallery)
- [x] The setup assistant (content/assistant)
- [x] The message pane (content/stub.xhtml & co)

Standard8:
#1333 is looking at removing jQuery from the gallery.

Standard8:
#1345 is the start of the rewrite for stub.xhtml. There's a long way to go, but it has started.

Standard8:
Status update: a few parts of the conversation view are now based on react - the conversation header and footer, some of the message header, and the attachments display (and a couple of other minor parts). These are getting things working again for 68, though there's a few bugs now.

That's given me enough proof of concept that I'm now working on a larger rewrite of the whole conversation view dropping jquery and replacing the templating mechanism we had there.

I'm not sure the eta for this work - I currently have a few basic items on the messages being displayed, but there's a lot more to do, will post an update or branch when I can.

Standard8:
I've just pushed the switch of the message list to React onto the master branch. It isn't quite complete yet (I'll file more issues in a moment), but it has now removed jQuery. Hence calling this done.

PR #1333, opened 2018-12-27
Replace the gallery with a React based implementation, dropping jQuery and lightbox.

Standard8:
Part of #1317. I'm currently testing this a little more and discussing with @protz, but I think this is a basically complete enough implementation that doesn't pull in too much more and transitions away from jQuery for the gallery.


Standard8:
@protz took a brief look and agrees this is better than jQuery implementation, hence landing this.

I will be looking at jsx generation soon.

PR #1345, opened 2019-02-19
Start Rewriting stub.xhtml with react

Standard8:
This is a WIP for converting stub.xhtml to React. This is so that we can replace jQuery (#1317).

I'm suspecting this is going to take a while to complete. I think we may want to introduce something like Redux as well, but I haven't investigated that yet.

At the moment I have the header, footer and attachment menu items moved across to React. Selecting different messages after the first one in a folder is broken though, as I've not yet fixed the parts for updating the conversation header (I think this is where redux may come in).

I might aim to get the first part landed with just the header/footer re-implemented and do the message list later. It might depend on how complicated it looks to do.


Standard8:
I've just pushed an update that introduces redux and gets most of the conversation header working properly, as well as restoring switching between messages.

With redux, we now have actions to update a store, which then updates the display. My general philosophy here is that modules that are updating the display (e.g. conversation.js) should dispatch an action rather than updating the html direct. This has the advantage of not reaching into the generated html, and for making it clearer where updates are coming from. It also allows for a possible future where the message pane could be treated as completely separate (e.g. separate process) and we just post messages between the two.

Likewise, I'm also trying to move anything that interacts with the core of Thunderbird (e.g. marking messages read), into the action handlers. Again, this helps to provide a clearer separation between layout and actual actions.

The other benefits I'm starting to see is to be able to move things like linkifySubject out of modules code and into the rendering direct. This should make it much easier to know how the messages are constructed. At some stage we may be able to put together some tests (either automated or manual) for this, to easily check various messages are rendered correctly.

I think once I've finished the conversation header part, I'll look towards merging this. The message list itself is likely to be a lot more complicated, so that'll be a reasonable separation.

Standard8:
Last update for today. I've re-arranged the top-level rendering so that we now have a `conversationWrapper`. I also adjusted the footer to use the new action architecture.

I might look at simplifying the action architecture a bit more next, though some of that may come later if we move some of the chrome interactions into jsms...

Standard8:
Ok, I've decided that what we have now is good enough to be merged into master. It seems to work and maintains functionality. There's possibly a little bit more overhead, but I think it is worth getting this in with incremental chunks.

PR #1372, opened 2019-10-12
LGTM warnings fixes

Hi-Angel:
This is essentially are fixes of some of the warnings [reported here](https://lgtm.com/projects/g/protz/thunderbird-conversations/alerts/?mode=tree).

Commit `content: expand menu in displayMenu` may need a bit of attention in case I'm missing something.

### DIFF
--- a/addon/content/messageHeaderOptions.jsx
+++ b/addon/content/messageHeaderOptions.jsx
@@ -130,7 +130,7 @@ class _MessageHeaderOptions extends React.PureComponent {
       document.addEventListener("keypress", this.keyListener);
     }
 
-    this.setState({expanded: !this.state.expanded});
+    this.setState(prevState => ({expanded: !prevState.expanded}));
   }
 
   clearMenu() {
--- a/addon/modules/message.js
+++ b/addon/modules/message.js
@@ -1132,7 +1132,7 @@ Message.prototype = {
   },
 
   set read(v) {
-    return msgHdrsMarkAsRead([this._msgHdr], v);
+    msgHdrsMarkAsRead([this._msgHdr], v);
   },
 
   get starred() {
@@ -1148,7 +1148,7 @@ Message.prototype = {
   },
 
   set tags(v) {
-    return msgHdrSetTags(this._msgHdr, v);
+    msgHdrSetTags(this._msgHdr, v);
   },
 
   get collapsed() {
--- a/addon/content/assistant/assistant.xhtml
+++ b/addon/content/assistant/assistant.xhtml
@@ -121,7 +121,7 @@
 
     <div class="buttonBox">
       <button onclick="onCustomSetup(event);">&assistant.review;</button>
-      <button onclick="onFinish(event);" id="applyButton" class="applyChanges">&assistant.apply;</button>
+      <button onclick="onFinish();" id="applyButton" class="applyChanges">&assistant.apply;</button>
     </div>
 
     <div id="more">


addon/content/js/handlebars-v3.0.0.js
```diff
@@ -209,6 +209,27 @@ var __module2__ = (function(__dependency1__, __dependency2__) {
     registerDefaultHelpers(this);
   }
 
+  var logger = {
```
Standard8: I don't want to modify this files - it is vendor imported at a specific version, and if anything we should be fixing upstream and updating it.

In this case, I'm working on a branch that will remove this completely that I hope to land in a month or two. Seeing as there's nothing critical here, I think we should just drop the changes to this file.

addon/content/messageHeaderOptions.jsx
```diff
@@ -130,7 +130,7 @@ class _MessageHeaderOptions extends React.PureComponent {
       document.addEventListener("keypress", this.keyListener);
     }
 
-    this.setState({expanded: !this.state.expanded});
+    this.setState({expanded: true});
```
Standard8: This is wrong, since we'll now never get collapsed. This should use the callback functionality suggested in the details of the alert.

addon/content/js/jquery.tokeninput.js
```diff
@@ -259,7 +259,7 @@ $.TokenList = function (input, settings) {
 
   // Pre-populate list if items exist
   function init_list () {
-    li_data = settings.prePopulate;
+    var li_data = settings.prePopulate;
```
Standard8: Similar to handlebars, this is also an imported vendor file that's going away soon. Lets just drop this change please.

addon/modules/message.js
```diff
@@ -1132,7 +1132,7 @@ Message.prototype = {
   },
 
   set read(v) {
-    return msgHdrsMarkAsRead([this._msgHdr], v);
+    msgHdrsMarkAsRead([this._msgHdr], v);
```
Standard8: TIL

addon/content/js/handlebars-v3.0.0.js
Hi-Angel: Thanks, done

addon/content/messageHeaderOptions.jsx
Hi-Angel: Thanks, I rewrote it

addon/content/js/jquery.tokeninput.js
Hi-Angel: Thanks, dropped

Hi-Angel:
While on it, I can as well ask it here: I've got a modified version of `package-lock.json`, apparently it was modified by some of `npm` commands. It now has a bunch of `"optional": true` lines added.

Is it an autogenerated code? If it isn't, then should I send these changes upstream as well?

For now I temporarily made git to ignore this file, so it wouldn't accidentally get into a commit.

addon/content/messageHeaderOptions.jsx
Hi-Angel: Oh, I figured I should rewrite the commit message, let me do it real quick

Hi-Angel: Done

Standard8:
> While on it, I can as well ask it here: I've got a modified version of `package-lock.json`, apparently it was modified by some of `npm` commands. It now has a bunch of `"optional": true` lines added.

This could be that you're running a newer version of npm than I am. The latest version of package-lock.json was generated on npm 6.5.0.

> Is it an autogenerated code? If it isn't, then should I send these changes upstream as well?

You should just be able to abandon the changes. Really I should change the documentation to suggestion using `npm ci` rather than `npm install` - The install option can sometimes grab the latest dependencies and rewrite the file to incorporate changes. The ci option uses the lock file to install exactly what's in package-lock.json and doesn't make any changes to it.

PR #1386, opened 2019-12-01
Switched to functional component in messageFooter

siefkenj:
I'm looking to clean up some of the react code and hopefully make Conversations more reliable (and testable?). Let me know if these contributions are welcomed :-)

### DIFF
--- a/addon/content/contactDetail.jsx
+++ b/addon/content/contactDetail.jsx
@@ -5,6 +5,27 @@
 /* globals PropTypes, React, ReactRedux, StringBundle */
 /* exported ContactDetail */
 
+/**
+ * Returns an SVG icon from `.../material-icons.svg#{hash}`
+ *
+ * @param {string} { type }
+ * @returnType {React.Node}
+ */
+function MaterialIcon({ type }) {
+  return (
+    <svg
+      className="icon"
+      viewBox="0 0 24 24"
+      xmlns="http://www.w3.org/2000/svg"
+      xmlnsXlink="http://www.w3.org/1999/xlink"
+    >
+      <use
+        xlinkHref={`chrome://conversations/skin/material-icons.svg#${type}`}
+      ></use>
+    </svg>
+  );
+}
+MaterialIcon.propTypes = { type: PropTypes.string };
 class _ContactDetail extends React.PureComponent {
   constructor(props) {
     super(props);
@@ -127,14 +148,7 @@ class _ContactDetail extends React.PureComponent {
                 title={this.strings.get("copyEmail")}
                 onClick={this.copyEmail}
               >
-                <svg
-                  className="icon"
-                  viewBox="0 0 24 24"
-                  xmlns="http://www.w3.org/2000/svg"
-                  xmlnsXlink="http://www.w3.org/1999/xlink"
-                >
-                  <use xlinkHref="chrome://conversations/skin/material-icons.svg#content_copy"></use>
-                </svg>
+                <MaterialIcon type="content_copy" />
               </button>
             </span>
           </div>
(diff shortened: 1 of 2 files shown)


siefkenj:
Thanks for the quick response :-).

Function components are the direction the react community is headed in. They eliminate the component lifecycle methods and replace them with [hooks](https://reactjs.org/docs/hooks-intro.html), which make a component much easier to think about.

The main downside of function components is that it's common to define inline functions, and if there are a lot of these, it's harder to read than a class with a lot of methods (though perhaps in both cases, the complexity of the component should be split...).

As for performance, a function component can be wrapped with `React.memo` (see [here](https://reactjs.org/docs/react-api.html#reactmemo)) which should produce equal or better performance to `React.PureComponent`. I guess the common practice is to build, profile, and then selectively optimize components, but I guess rect dev tools doesn't run in Thunderbird.

Which, brings me to an idea: If a lot of the GUI could be separated from direct Thunderbird calls, then a mock-thunderbird could be created for the browser and sections of the GUI could be run directly in the browser (with react and redux devtools). That should make development of GUI portions easier (browser refresh instead of thunderbird restart (unless there's a way to reload the extension without restarting Thunderbird)) and it would also allow for the many browser-based testing frameworks to be run along with their profiling tools, etc..

I did have a few questions:

  * I see the redux store is created multiple times during a Thunderbird lifecycle. When exactly is it created, and is there a reason you didn't opt for creating it once and reusing it from then on?
  * I see that many reducer actions have side effects. It's tradition to define sideeffects inside `actions` and have the reducers be side-effect free. Was this a purposeful design choice?
  * How do you feel about ES modules/bundling? Do you want to keep all the code native, or would you be happy if webpack were used?
  * Are you familiar with [Redux Toolkit](https://github.com/reduxjs/redux-toolkit)? I think it would allow for simpler reducers and to split off sideffects without introducing too much boilerplate.

Standard8:
Thank you for the links and comments. I'll need a week or two to be able to get time to look into these and digest and understand what's going on.

I can comment on a couple of things now though.

The new code has been designed so that the reducer actions will be the only things communicating with the core of the add-on/Thunderbird. There were two ideas here. One was so that we had a clear location of all the calls going to/from the backend. The other was that eventually we might be able to migrate to a system whereby the entire message display is in an unprivileged sandbox with message passing to/from the backend. I'm not sure how realistic that is, and it may need changes to Thunderbird, but overall it seems a much better structure.

This separation would indeed most likely allow the backend to be stubbed out, with the UI to be displayed as a normal page. You could likely do this within Thunderbird as there's content tabs. You might be able to use the react developer tools - The Firefox developer tools can be used for Thunderbird as well (Tools -> Developer Tools -> Developer Toolbox).

> I see the redux store is created multiple times during a Thunderbird lifecycle. When exactly is it created, and is there a reason you didn't opt for creating it once and reusing it from then on?

The message pane is used for displaying lots of different things, and Thunderbird is in control of it. So sometimes Thunderbird will wipe it out completely, as we switch to a different folder. We basically create the store when the page is loaded into the message pane, then (hopefully) we are keeping the same store for the lifetime of that page load.

Unfortunately we can't do much about Thunderbird's control paths here, at least not in the short term.

> I see that many reducer actions have side effects. It's tradition to define sideeffects inside actions and have the reducers be side-effect free. Was this a purposeful design choice?

The reducer/action design isn't ideal, and was more put together to get something working with the aim of tidying it up later. The rewrite has already taken long enough and it isn't quite complete yet (enigmail, quick reply). I'm open to changes where they make sense / don't cause a lot of extra boiler plate.

> How do you feel about ES modules/bundling? Do you want to keep all the code native, or would you be happy if webpack were used?

Bundling I don't want to do at the moment, for one thing it complicates the submission process for the add-on to addons.thunderbird.net.

ES Modules I'd like to do at some stage, again, I was mainly getting the basic rewrite done first and tidying up what was there. The good news is that Thunderbird should support these natively, though I'm not quite sure how it combines with the special context that the add-on runs within.

Again, the main reason I hadn't experimented here was because of the rewrite and the earlier code used the xpcom import mechanisms which aren't compatible with modules. Hopefully most of that has gone away now - apart from the bits in stub.js and reducer.js, which we'd probably need to figure out somehow.

siefkenj:
This change will be incorporated in another PR with the `<Icon />` split into a module for reuse. Closing for now.

PR #1387, opened 2019-12-04
Change options.html to React/Redux

siefkenj:
This PR changes `options.html` to use React/Redux. It also serves as an example of how to make components that can be developed/debugged in the browser.

An `es-modules` directory was created that contains `ui.js`, which exports React and friends (which are presumed to be loaded as global variables) and `thunderbird-compat.js`, which mocks the Thunderbird API calls needed to read and set preferences. If the native `browser` object is detected, `thunderbird-compat.js` will do nothing.

There's a new build script, `build-dev-html.sh` which will build and serve the project on `localhost:8126`. Navigating to http://localhost:8126/options.html you can see `options.html` rendered in the browser and can use the React/Redux dev tools to debug it. Additionally, all files are watched for changes and automatically recopied/recompiled. Separating components enough that they can be run in the browser should allow for easier testing of individual components.

This PR also includes Redux Toolkit (hopefully this is compatible with AMO!) which simplifies Redux development.


Standard8:
Hi, I'm sorry I've not got to this yet. I'm hoping to get time on Thursday.

siefkenj:
No worries. I've been piling on the code, unsolicited :-P.

siefkenj:
Have you had any time to look at this? I've got some other ideas that build on it.

addon/options.html
```diff
@@ -6,107 +6,26 @@
   <head>
     <meta charset="utf-8">
     <link rel="stylesheet" href="options.css" type="text/css">
+    <!-- React and friends must be loaded before the ui.js module -->
+    <script type="text/javascript"
+      src="./content/vendor/react.js"></script>
```
Standard8: Are we not able to import these as modules? I'm wondering if we could actually get to the stage where options.js is the only script statement here, and then everything is imported into that via import statements?

This would certainly be nicer from the viewpoint of handling globals.

package.json
```diff
@@ -14,6 +14,7 @@
     "thunderbird": ">=68.0a1"
   },
   "dependencies": {
+    "@reduxjs/toolkit": "^1.1.0",
```
Standard8: Generally I prefer exact versions here so that we know when these are getting bumped.

package.json
```diff
@@ -35,10 +36,13 @@
...
     "npm-run-all": "4.1.5",
     "prettier": "1.19.1",
-    "prop-types": "15.7.2"
+    "prop-types": "15.7.2",
+    "cpx": "^1.5.0",
+    "serve": "^11.2.0"
   },
   "scripts": {
     "build": "./build.sh",
+    "build:dev-html": "./build-dev-html.sh",
```
Standard8: I know the main build script isn't there yet, but it'd be nice to start putting these in scripts/. Maybe call this one "dev" - something that's simpler to type.

README.md
```diff
@@ -33,6 +33,18 @@ Building
 
 This will package a `converstions.xpi` file of the latest codebase which can be installed via add-on manager in Thunderbird (hint: you can drag & drop it onto the add-on manager view).
 
+Development
```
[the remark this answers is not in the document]
siefkenj: Makes sense. That is probably best to do on the first PR that adds tests.

addon/content/es-modules/thunderbird-compat.js
```diff
@@ -0,0 +1,219 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+// A compatability layer that can be imported whether in the browser or
+// in Thunderbird
+
+const DEFAULT_LOCALE = {
```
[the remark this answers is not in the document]
siefkenj: I agree. I see two solutions: either use `fetch` to get them, in which case their availability becomes async, or have webpack etc. compile them in for dev mode. I just stuck the English in so everything was syncronous, but I'll see if I can get something else working.

addon/content/es-modules/thunderbird-compat.js
```diff
@@ -0,0 +1,219 @@
...
+    description: "Options - title for monospaced sender setting.",
+  },
+
+  "options.monospaced_senders_desc": {
+    message:
+      "Messages from those addresses/mailing-lists should be displayed in monospaced fonts (separate using commas).",
+    description: "Options - description for monospaced sender setting.",
+  },
+};
+const DEFAULT_PREFS = {
```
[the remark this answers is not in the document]
siefkenj: Where are the default prefs located? The trouble is for es modules, one must use the export keyword (or store something as a global) and the export keyword breaks normal scripts. Maybe it can be put in a try-catch statement...? But I think this is forbidden since exports need to be statically analyzable.

addon/options.html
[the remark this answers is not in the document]
siefkenj: Again, to import as a module, one needs to use the `export` keyword, which breaks regular JS. Once all code is switched to modules, this wouldn't be a problem, but for now....

So, I basically see two options: 1) have the js included in `options.html` or have a module included in `stub.html` which sets React etc. as globals to make the existing code work.

package.json
siefkenj: I'll change this :-)

package.json
siefkenj: Sure. I can move the scripts. I called it `dev-html` because it only builds a browser-based version. I thought `dev` would be confusing, since nothing for thunderbird is actually produced...

addon/content/es-modules/thunderbird-compat.js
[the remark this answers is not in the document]
siefkenj: I found `prefs.js`. I'm a little confused about what's going on in that file...should I just import the `Prefs` object and instantiate a new copy? It looks like there's automatic migration stuff that goes on. Did we want that triggered in the dev version?

addon/options.html
siefkenj: Done

siefkenj:
I found a way to hack around the UMD module definitions by adding `modules-compat.js`. Hopefully my workaround makes sense!

Standard8:
> I found a way to hack around the UMD module definitions by adding `modules-compat.js`. Now react, etc can be imported by the module system. Hopefully my workaround makes sense!

Ah interesting, I hadn't realised that React didn't support that yet, I just found https://github.com/facebook/react/issues/14635 and https://github.com/facebook/react/pull/15037 that will hopefully mean they'll support it sometime soon. In the meantime, the work around is probably good enough.

I'll try and take a look at the rest of this some time this week.

README.md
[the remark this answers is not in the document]
Standard8: Filed #1394 for tracking.

package.json
Standard8: Hmm, how about just `dev-html` then and skip the `build:` prefix?

siefkenj: Sound good.

siefkenj:
Any updates on the merge status of this? I almost have a PR that implements tests. It builds on this one and would probably be easier to understand once this PR is merged. I will be away from computers Dec 24-Jan 5.

package.json
```diff
@@ -35,10 +36,13 @@
     "eslint-plugin-react": "7.16.0",
     "npm-run-all": "4.1.5",
     "prettier": "1.19.1",
-    "prop-types": "15.7.2"
+    "prop-types": "15.7.2",
+    "cpx": "1.5.0",
```
Standard8: Can we use cpx2 please? That seems to be the supported version now based on https://github.com/mysticatea/cpx/pull/49 (there's also a minor security issue in cpx according to `npm audit`.

scripts/build-dev-html.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/sh
+NOT='.sh$|^.git|.jsx$|^tests|^.eslint|^.travis|^package.json$|^package-lock.json$'
```
Standard8: I feel like we should harmonise these scripts a bit, but this will do for now.

scripts/build-dev-html.sh
```diff
@@ -0,0 +1,46 @@
...
+done
+
+echo "    Copying non-jsx files"
+npx cpx "**/*.{js,html,xhtml,css,svg,png,gif,ico,dtd,properties,json}" ../${DIST}/ --watch --verbose &
+
+popd
+
+# This is a bit of a hack, but since all processes are being backgrounded, if
+# we don't wait for them to finish, our user message will get burried
+sleep 4
```
Standard8: Can't we just `wait $(jobs -p)` here?

addon/options.jsx
```diff
@@ -0,0 +1,419 @@
...
+}
+
+//
+// React components to render the options types
+//
+
+function ChoiceOption({ title, desc, choices = [], value, name, onChange }) {
+  const elementName = `choice_${title}`.replace(/\s+/g, "");
+  return (
+    <>
```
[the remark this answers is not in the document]
siefkenj: How about `<React.Fragment>` ?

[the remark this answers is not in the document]
siefkenj: If they were made `<div>` it would mess with the grid layout.

scripts/build-dev-html.sh
siefkenj: I reorganized this in the testing PR. The trouble is, when things are launched with `--watch`, we can't `wait` them, because they don't terminate.

addon/content/es-modules/modules-compat.js
```diff
@@ -0,0 +1,40 @@
...
+// module is loaded may be assumed to be AMD modules. It is your responsibility
+// to handle the initialization of the modules.
+
+export const amdModules = [];
+
+// eslint-disalbe-next-line
+window.define = function define(deps, moduleFactory = deps) {
+  // Sometimes the function is called `define(f)` and sometimes
+  // it's called `define([],f)` when there are no deps. Either way,
+  // normailze the result.
```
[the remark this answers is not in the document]
siefkenj: I will try to find a way to turn on a spellchecker for comments...

scripts/build.sh
```diff
@@ -8,23 +8,22 @@ rm -rf $DIST
...
+cp "node_modules/@reduxjs/toolkit/dist/redux-toolkit.umd.js" $VENDOR_DIR
 
 cp LICENSE README.md $DIST/
 
 pushd $ADDON_DIR
 
 for a in $(git ls-files | grep '.jsx$'); do
   echo $a
-  babel --config-file=../babel.config.js $a --out-dir ../$DIST/$(dirname $a)
+  babel --config-file=../babel.config.js $a --out-dir ../$DIST/$(dirname $a) &
```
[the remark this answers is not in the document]
siefkenj: Yeah. It speeds up builds by quite a bit! I wonder at what point the project should move to an actual build system...

package.json
siefkenj: Good thinking. I didn't know `cpx2` was a thing!

scripts/build-dev-html.sh
Standard8: Ah ok, I missed that. No problem then.

Standard8:
@siefkenj I've just pushed https://github.com/protz/thunderbird-conversations/commit/a8294151564a407a9173df2b8ff07f521e2fa2ac which updates the React modules to the latest version, which includes the UMD fix https://github.com/facebook/react/pull/15037 - so I think we might be able to simplify some of these imports now?

Though not entirely sure if we need to wait for the other modules to be updated as well.

siefkenj:
I think it's easiest to wait for all modules to get updated to accept native loading.

PR #1397, opened 2019-12-21
Add `jest` tests for options.html

siefkenj:
This PR adds `jest` tests for the `options.html` page. The tests test the return types of each of the option components, the actions which trigger the saving of options, and the complete page where a "click" is simulated to test that all components are connected to actions appropriately.

These tests can serve as a template for how to move more of Conversations under test coverage.

Some odd hacks were required to get the code running under both browser/thunderbird and node. Specifically, the browser uses native ESM module loading while node uses CJS module loading. Most of the dependencies test for CJS module loading first and use that if available. This breaks some ESM module loading as a consequence. The workaround is as follows:
  * Use the `esm` package for node to load scripts an emulate native module loading
  * Prepend vendor files with a script that disables the `global.require` object, so that node-style module loading is avoided.
  * Detect Node in `ui.js`. If Node is detected, use CJS module loading for React and friends. This is needed because the testing library also loads React using CJS module loading, and having two copies of React messes things up.


Standard8:
I'll have to look at this after Christmas, unfortunately I've run out of time now.

siefkenj:
No worries. I'm on holiday too! Thanks for your careful reviews :-)

addon/__test__/options.test.jsx
```diff
@@ -0,0 +1,191 @@
+/* eslint-disable */
```
Standard8: Can we put the tests in either `addon/test` or maybe `test` in the top level-directory? I think I'd prefer that to having the underscores.

Also, I'd like ESLint enabled here if possible - when we've been enabling it in Firefox we found lots of places where tests weren't running correctly or as expected. Turning ESLint on helped to show those up.

addon/__test__/options.test.jsx
```diff
@@ -0,0 +1,191 @@
...
+  const ROOT_PATH = path.join(__dirname, "..");
+  const filePath = path.join(ROOT_PATH, url);
+
+  const data = fileSystem.readFileSync(filePath, "utf8");
+  return Promise.resolve({
+    json: function() {
+      return Promise.resolve(JSON.parse(data));
+    },
+  });
+};
```
Standard8: Just a thought: is there any way we can avoid the global setup duplication for each test file?

Are we really needing React to be able to be loaded as es6 files?

scripts/build-dev-html.sh
```diff
@@ -1,46 +1,51 @@
-#!/bin/sh
+#!/bin/bash
 NOT='.sh$|^.git|.jsx$|^tests|^.eslint|^.travis|^package.json$|^package-lock.json$'
 DIST=dist-dev-html
 VENDOR_DIR=$DIST/content/vendor
 ADDON_DIR=addon
 DEV_SERVER_PORT=8126
 
+WATCH_FLAG=""
+if [ $1 = "--watch" ]
```
Standard8: On bash on Mac, I'm seeing `scripts/build-dev-html.sh: line 9: [: =: unary operator expected`

addon/__test__/options.test.jsx
siefkenj: By default Jest scans for `__test__` directories. I believe this can be changed by adding a jest config file, but it would be non-standard. Would you like me to look into that?

addon/__test__/options.test.jsx
siefkenj: I can move these to an import that is imported to all tests. Since there was only one test at this point, I put them at the top.

scripts/build-dev-html.sh
siefkenj: Ah, apparently, it should be `"$1"` to work on older BASH versions.

siefkenj:
I think I've addressed all comments. It was difficult to figure out how to resolve the submodule nonsense, but I finally got it!

siefkenj:
Do you have time to give this another lookover?

Standard8:
Yes, sorry I was hoping to today, but my normal computer was stolen by my son. Will definitely have a look tomorrow.

PR #1400, opened 2019-12-28
Fixes #1377: Reimplement Enigmail Integration

### LINKED ISSUE #1377: Latest master: fix Enigmail integration
I've just pushed the message list transition of React to the master branch.

This has switched us away from jQuery.

At the moment, the Enigmail integration is not present. That needs adding back before we can ship it.

Standard8:
Fixes #1377 

This is a WIP branch that is for reimplementing Enigmail integration.

At the moment, this does:

* Various cleanup
* Gets signed & decrypted indications working, including the tooltip & clicking on the indication.
* Selecting information on different messages in the thread should also work.

One thing I've also done is to remove the notification bar that was present for signed indications only (afaict). The functionality seems to be duplicated with the signed message tag.


Standard8:
Just to note, I've added decryption indication handling now as well. Still need to handle multiple messages in a thread.

Standard8:
Handling switching between messages in the same thread was easier than expected, so I've just pushed that and tidy ups to the branch. Will give this a run through later today, but hopefully looking good so far.

Standard8:
I think we're good to go now, at least to get some wider testing.

PR #1401, opened 2019-12-30
readme: replace two git commands with an analogous one

Hi-Angel:


Standard8:
Thanks, didn't know you could do that.

PR #1405, opened 2020-01-03
Change Babel configuration to work for just Firefox 68 or later, since that is what we currently support.

Standard8:
The 'esmodules' definition equates to Firefox 60 or later, [see here](https://github.com/babel/babel/blob/9febf6388233c3d23d4fa9deca5ca8f682567573/packages/babel-preset-env/data/built-in-modules.json).

This currently means the generated code has changes such as async/await swapped into generators, arrow functions replaced and more.

Whilst this probably doesn't matter too much, I think I'd rather keep the changes minimal and let the generated code use all the features of js that are available in the minimum version that we support (68). There's would be a slight reduction in file size, maybe some very small performance advantages as well (due to not having polyfill functions).

@siefkenj can you see any reason not to do this?

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -5,7 +5,10 @@ module.exports = function(api) {
     [
       "@babel/preset-env",
       // We don't want es modules to be bundled; we'll use native loading!
-      { targets: { esmodules: true }, modules: false },
+      {
+        targets: { browsers: "Firefox >= 68.0" },
+        modules: false,
+      },
     ],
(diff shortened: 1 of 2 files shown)


siefkenj:
I think this is a good change. Native features should give us better performance in the long run and easier debugging in the short run.

Standard8:
Thanks. I'm hoping to get time to look at your PR later today.

Issue #1428, opened 2020-01-16
Individual messages of a thread are not focused on in message preview pane


fridgelord:
* **Thunderbird version:** 68.4.1
* **Thunderbird Conversation version:** 3.0.10
* **Operating System:** Windows 10

**What did you do?**
Click on a message of a **collapsed** thread in message preview window.

**What did you expect to happen?**
Before the update individual message could be highlighted (with a slight shade around the message frame) and focused on by clicking on it in the message preview pane and you could use keyboard shortcuts on this particular message. For example CTRL+(SHIFT)+R would reply to this message (this might be related to Issue #1399, it is needed not only to focus on the message preview pane, but also on particular message). F & G would move between messages in the thread. It was possible to tag single message using numbers.

**What actually happened?**
If thread is collapsed, CTRL+(SHIFT)+R has no effect (unless there is only one message in the thread), F & G move between threads and not messages. Tagging with numbers tags all messages in thread.

Standard8:
Thank you for the explanation. I must admit I didn't fully understand some of that code or what it did, which is why it hasn't been re-implemented.

I think it makes sense to have this, so I'll mark it as a bug, and hopefully someone can find time to get to it soon.

Standard8:
@siefkenj I was wondering if you might like to take this one.

I think the old handlers for this were in message.js on the 2.14.x branch: https://github.com/protz/thunderbird-conversations/blob/2.14%2B/modules/message.js (look for `KeyListener`).

We could maybe just implement one part first, e.g. reply/forward and then think about the rest.

Standard8:
I believe this is now fixed by #1432.

PR #1432, opened 2020-01-20
Added keyboard navigation

siefkenj:
Add basic keyboard navigation to the conversations view. This address issue #1428

### DIFF
--- a/addon/content/message.jsx
+++ b/addon/content/message.jsx
@@ -7,6 +7,9 @@
            MessageDetails, MessageNotification */
 /* exported Message */
 
+const { isAccel } = ChromeUtils.import(
+  "chrome://conversations/content/modules/stdlib/misc.js"
+);
 class Message extends React.PureComponent {
   constructor(props) {
     super(props);
@@ -14,12 +17,10 @@ class Message extends React.PureComponent {
       "chrome://conversations/locale/template.properties"
     );
     this.onSelected = this.onSelected.bind(this);
+    this.onKeyDown = this.onKeyDown.bind(this);
   }
 
   componentDidMount() {
-    this.li.addEventListener("focus", this.onSelected, true);
-    this.li.addEventListener("click", this.onSelected, true);
-    this.li.addEventListener("keydown", this.onSelected, true);
     if (
       this.lastScrolledMsgUri != this.props.message.msgUri &&
       this.props.message.scrollTo
@@ -62,9 +63,6 @@ class Message extends React.PureComponent {
   }
 
   componentWillUnmount() {
-    this.li.removeEventListener("focus", this.onSelected, true);
-    this.li.removeEventListener("click", this.onSelected, true);
-    this.li.removeEventListener("keydown", this.onSelected, true);
     this.removeScrollListener();
   }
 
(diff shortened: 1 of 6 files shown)


addon/content/message.jsx
```diff
@@ -7,19 +7,20 @@
            MessageDetails, MessageNotification */
 /* exported Message */
 
+const { isAccel } = ChromeUtils.import(
```
Standard8: At some stage we're not going to be able to import scripts from the back-end into message content, so I think it'd be nice if we start working out how we're going to handle these kind of simple functions. 

Maybe we should just duplicate it for now, as far as I know there's nothing in general html/javascript land that will help us here, unless there's something React has?

addon/content/reducer.js
```diff
@@ -181,6 +190,29 @@ function messages(state = initialMessages, action) {
       MessageUtils.setTags(action.msgUri, action.tags);
       return state;
     }
+    case "MSG_TOGGLE_TAG_BY_INDEX": {
+      // If we're here we've been passed the index of a tag to toggle on the
+      // given message as well as the existing tags.
+      const allTags = MailServices.tags.getAllTags({}).map(normalizeTag);
```
Standard8: We could probably put this part into a utility function within MessageUtils. Generally, I'd like us to be working towards making messages handled in the reducer with simple calls through to the backend. The idea being that at some stage in the future, we can replace the direct calls with a message passing boundary, which we're likely to need as we complete the move to WebExtensions.

addon/content/message.jsx
siefkenj: There's `window.navigator.platform`. We could just unconditionally listen for ctrl or cmd (I don't use a Mac, so I don't know if this would cause surprises or not...).

In the long term, the Redux store should store the OS and the `isAccel` function should be imported as an es import. I can't figure out how to do es imports in the `stub.html` code, though.

Standard8: Ah yes we could use `navigator.platform`. I think we could just duplicate isAccel here, it isn't really worth a utility function on its own IMO.

At the moment we probably can't do ES imports, I suspect we're missing bits from the Thunderbird core to be able to do that. Though it might be possible to use if the original javascript file has type="module" in the script tag in the html file, but we probably can't move everything across yet, because of all the ChromeUtils.import calls.

siefkenj: We can't do imports in `xhtml`, it seems...it has to be regular `html`...

siefkenj:
Any chance at another review?

addon/content/reducer.js
```diff
@@ -33,6 +33,10 @@ const initialSummary = {
   subject: "",
 };
 
+// conversations sometimes stores a tag as
+// { color, id, name } but the internal TB
+// representation is { color, key, tag }.
+// This function converts to the conversations representation
```
Standard8: This function (`modifyOnlyMsg`) handles only updating a single message in the reducer. I'm not sure where this comment is meant to be attached to.

Standard8:
@siefkenj So I was just playing with this before heading to bed, and noticed an issue. I'm pretty sure it is this PR as I applied the old one and it went away.
 
The issue is that when you hover over the name of someone (author/to/cc etc), the contact display fades into view. If you then try clicking one of the buttons, the pop-up gets shifted and the click doesn't happen.

We had this before, but it magically disappeared at some stage, so I never worked out what it was.

addon/content/reducer.js
siefkenj: My bad...that leaked from my PR for the tag refactoring...

siefkenj:
Very interesting. I'll try to reproduce. The key change in this PR is setting the `tabIndex` (otherwise events never made it to the react component). I wonder if that messed something else up...

Standard8:
I have just remembered there was some strange `tabIndex` handling going on with the pre-react code that I didn't bring across. There's a few links here, I don't know if they'll be useful or not:

https://github.com/protz/thunderbird-conversations/blob/b7753c6ccd34736b505f63384258cea5bacaaf02/content/stub.xhtml#L249
https://github.com/protz/thunderbird-conversations/blob/b7753c6ccd34736b505f63384258cea5bacaaf02/modules/message.js#L201
https://github.com/protz/thunderbird-conversations/blob/2.14%2B/modules/conversation.js

siefkenj:
I just tried to reproduce on a fresh profile and the buttons in the hover popup work just fine for me...

siefkenj:
Indeed it is the `tabIndex` that is causing the issue. If I remove the `tabIndex` in the inspector, everything works.

What appears to be happening is that the position parent (as exposed by `offsetParent` changes from `<body>` to the surrounding `<li>` when something with a `tabIndex` property is focused. I do not know why this is happening. Don't have access to Firefox 68, but it seems to work properly in 72: https://jsfiddle.net/k28ocg3a/

The easiest fix at this point seems to be to use a React portal to reparent the popup to `<body>`...

Standard8:
Thanks for finding that portal. I've just given it a brief one-minute test and it seems better. I'll give it a go over the next day and se ehow it runs.

PR #1433, opened 2020-01-20
Initial Cleanup of Tags

siefkenj:
Move dispatching out of `<MessageTag />` to allow for easier testing in the future. Converted `<MessageTag />` and `<MessageTags />` to functional components.


siefkenj:
I started cleaning up `<SpecialMessageTag />`, but I'm not sure what they are. They also don't seem to be used correctly. The react component lists `props.specialTags` as required, but that is not passed in in `messageHeader.jsx`. The code also really confuses me. It appears the `dispatch` gets passed a function which gets passed somewhere else. `dispatch` should only send serializable data...

Standard8:
I guess I wrote all that code a bit too quick from a quick look around, and didn't document enough.

When Conversations isn't installed, some add-ons display tags and notifications in the message window, e.g. [DKIM Verifier](https://addons.thunderbird.net/thunderbird/addon/dkim-verifier/) and [Enigmail](https://addons.thunderbird.net/thunderbird/addon/enigmail/).

For Conversations, we provide "plugins" for these to adapt to our message display (see `addon/content/modules/plugins`). I've created newer functions where plugins can add special tags and notifications to a message. The special tags are tags that are not saved with the message, nor on the server, but are the result of these add-ons providing hints on the display.

> I started cleaning up `<SpecialMessageTag />`, but I'm not sure what they are. They also don't seem to be used correctly. The react component lists `props.specialTags` as required, but that is not passed in in `messageHeader.jsx`.

That's a mistake. The tags are shown when the message is expanded, via `message.jsx`, and when the message is collapsed, via `messageHeader.jsx`. I did the special tags over a few days, so I probably forgot to check that collapsing worked properly.

Note that I think the order difference is intentional between the two - for the collapsed view you more want to know the tags that are applied first, the signing/verified status are less important, when it is expanded, these don't want to be lost.

> The code also really confuses me. It appears the `dispatch` gets passed a function which gets passed somewhere else. `dispatch` should only send serializable data...

I'm assuming you're referring to `onClick`. If so that isn't a function, just badly named. What happens is that a plugin adding a [message tag adds an onClick object](https://github.com/protz/thunderbird-conversations/blob/1cc66afc66c61771b9baf47785773b2ed60b9159/addon/content/modules/plugins/enigmail.js#L675) with `type` and `detail` strings.

When the tag is clicked, this [detail is passed back](https://github.com/protz/thunderbird-conversations/blob/1cc66afc66c61771b9baf47785773b2ed60b9159/addon/content/messageTags.jsx#L112) and the [plugin can act appropriately](https://github.com/protz/thunderbird-conversations/blob/1cc66afc66c61771b9baf47785773b2ed60b9159/addon/content/modules/plugins/enigmail.js#L738).

I'm quite happy for `onClick` to be renamed here, it could be made optional as well - then `canClick` could also be replaced by if the new `onClick` exists or not.

Thanks for highlighting these. I obviously didn't finish that part off enough.

siefkenj:
Thanks for the explanation. Do could you provide some sample data for `<SpecialTags />`? I don't use either of those additional extensions, so it's hard for me to test.

Standard8:
DKIM Verifier requires no set-up and gives an indication for most messages. That's probably the easiest way at the moment, although I could try and grab some example data if that'll help.

Also, I've just pushed a change that fixes most of the current React warnings that we see about required props and props of the wrong types. I've left the tag ones alone as you're working on them :-)

siefkenj:
I just tried to install DKIM, but it won't install...It would be helpful if you could provide some sample data.

siefkenj:
Okay. I've moved `dispatch` calls out of the `*Tags` elements. They now trigger callbacks and are dispatched by the parent. This should make them more testable. It also eliminates several props that were being passed down merely to be used for a dispatch.

Standard8:
@siefkenj I'll try and remember to get a DKIM dump for you tomorrow for this.

Standard8:
DKIM Dump:

```
{
  "canClick": false,
  "classNames": "dkim-signed  SUCCESS",
  "icon": "chrome://conversations/skin/material-icons.svg#edit",
  "name": "DKIM signed",
  "tooltip": {
    "type": "dkim",
    "strings": [
      "Valid (Signed by example.com)",
      []
    ]
  }
}
```

```
{
  "canClick": false,
  "classNames": "dkim-signed warnings SUCCESS",
  "icon": "chrome://conversations/skin/material-icons.svg#edit",
  "name": "DKIM signed",
  "tooltip": {
    "type": "dkim",
    "strings": [
      "Valid (Signed by example.com)",
      [
        "Insecure signature algorithm"
      ]
    ]
  }
}
```

siefkenj:
Okay. I've tested the `SpecialTags` and they seem to work :-). This is ready for review.

addon/content/reducer.js
```diff
@@ -396,7 +396,8 @@ function messages(state = initialMessages, action) {
     }
     case "TAG_CLICK": {
       const msg = Conversations.currentConversation.getMessage(action.msgUri);
-      msg.msgPluginTagClick(topMail3Pane(window), action.event, action.detail);
+      const { event, details } = action;
+      msg.msgPluginTagClick(topMail3Pane(window), event, details);
```
Standard8: I'm not quite sure we need this breakdown, if it was more fields then maybe. However, this feels a little unnecessary.

addon/content/messageTags.jsx
```diff
@@ -54,170 +45,156 @@ MessageTag.propTypes = {
...
+
+  let folderItem = null;
+  if (folderName && !inView) {
+    folderItem = (
+      <li
+        className="in-folder"
+        onClick={onFolderClick}
+        title={strings.get("jumpToFolder")}
+      >
+        {strings.get("inFolder", [this.props.folderName])}
```
Standard8: `this.props.folderName` needs to be just `folderName`

addon/content/messageTags.jsx
```diff
@@ -54,170 +45,156 @@ MessageTag.propTypes = {
...
-      </ul>
+function SpecialMessageTags(props) {
+  const {
+    onTagClick,
+    onFolderClick = null,
+    specialTags,
+    strings,
+    inView,
+    folderName,
+  } = props;
```
Standard8: I just realised, how about we simplify these and do:

```
function SpecialMessageTags({
  onTagClick,
  onFolderClick = null,
  specialTags,
  strings,
  inView,
  folderName,
}) {
```
Then all the properties are documented directly in the function definition, and we don't need the separate breakdown.

siefkenj: Sure, I can change this if you like. The reason I destructure the `props` instead of doing it directly in the function definition is that I've been bit in the past by error messages. `"props" is undefined` is a lot faster for me to recognize than `TypeError: (destructured parameter) is undefined`.

PR #1445, opened 2020-02-22
Create a browser mapping for stub.xhtml to use to simulate WebExtensions

Standard8:
I realised that by exposing some extra APIs and doing a little extra work, that we could expose `browser` object to stub.xhtml that simulates the `browser` object that WebExtensions have.

We do this by registering the WebExtension part with an experimental API event, which then is used as a callback mechanism, so that calls to the `browser` object in stub.xhtml call back to the real `browser` object.

My intention is to allow us to start moving things in stub.xhtml across to use WebExtension APIs "directly". Hence we can progressively move across the direct calls to use WebExtension APIs - this should make it easier once we get to the final transition stage of moving the whole page to be a webExtension page (hopefully that'll work).

@siefkenj I thought you might be interested in this, as I think this lays the groundwork for things later. I'm planning on doing a little bit more work this weekend - to move the obvious create tab calls across and do a bit more work on the contacts side - before I land this.

### DIFF
--- a/addon/content/modules/conversation.js
+++ b/addon/content/modules/conversation.js
@@ -183,7 +183,30 @@ ViewWrapper.prototype = {
   },
 };
 
+// This is a workaround whilst we still have stub.xhtml being loaded in the
+// privileged scope. _ConversationUtils.getBrowser() simulates APIs and passes
+// them back to the webExtension process for handling by the real APIs.
+const SUPPORTED_BASE_APIS = ["tabs"];
+
 class _ConversationUtils {
+  setBrowserListener(listener) {
+    this._browserListener = listener;
+  }
+
+  getBrowser() {
+    const browser = {};
+    const self = this;
+    for (const api of SUPPORTED_BASE_APIS) {
+      const subApiHandler = {
+        get(obj, prop) {
+          return self._browserListener.bind(null, api, prop);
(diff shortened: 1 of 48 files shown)


Standard8:
The only thing we have to be aware of is that we'll have to add any new APIs we use to `SUPPORTED_BASE_APIS` in `conversation.js`. Unfortunately I couldn't think of an easy way around that.

addon/content/modules/conversation.js
```diff
@@ -183,7 +183,33 @@ ViewWrapper.prototype = {
...
+};
+
 class _ConversationUtils {
+  setBrowserListener(listener) {
+    this._browserListener = listener;
+  }
+
+  getBrowser() {
+    const browser = {};
+    for (const [api, subApis] of Object.entries(SUPPORTED_BASE_APIS)) {
```
siefkenj: A fancy alternative would be to use a `Proxy` for this. That way you could capture the attribute access and wouldn't need the `SUPPORTED_BASE_APIS`.

Standard8: Thanks, I've used one proxy so far, which gets around having the list of sub apis. Not quite sure it is worth going the double-stacking, but at least for now, this should be a bit easier.

PR #1446, opened 2020-02-23
Add basic tests for MessageTags

siefkenj:
This PR builds off of #1433 by adding basic Jest tests for `MessageTags` and `SpecialMessageTags`.

Because we can't use es-modules inside `stubs.xhtml`, a workaround involving exporting globals was needed to "simulate" the module system. When `stubs.xhtml` is moved into a WebExtensions context, then es-module loading should work and the hacks can be deleted.

Specifically, this PR stubs out `content/es-modules/components/message-tags.js`. `messageTags.jsx` should eventually be migrated to this file.


addon/content/es-modules/components/message-tags.js
```diff
@@ -0,0 +1,31 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/*
+ * This is a temporary wrapper. XUL doesn't allow es-modules, but eventually
+ * all components should be exported as es-modules. This code "re-exports"
+ * globals as es-modules, which enables them to be imported (e.g. for tests)
+ * or used as globals.
```
Standard8:
```suggestion
 * This is a temporary wrapper. Privileged code doesn't allow es-modules, but
 * eventually all components should be exported as es-modules. This code 
 * "re-exports" globals as es-modules, which enables them to be imported
 * (e.g. for tests) or used as globals.
```

addon/content/es-modules/components/message-tags.js
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * When the switch to a WebExtension is done, the actual code should be migrated
+ * here and the global variable workarounds should be removed.
+ */
+
+/* globals require */
```
Standard8:
```suggestion
/* eslint-env node */
```

addon/content/es-modules/components/message-tags.js
```diff
@@ -0,0 +1,31 @@
...
+
+// Make sure all the libraries that need to be in global scope are in global scope.
+import { React, ReactDOM, Redux, ReactRedux, RTK, PropTypes } from "../ui.js";
+Object.assign(window, { React, ReactDOM, Redux, ReactRedux, RTK, PropTypes });
+
+// Set up an object for the make-shift module emulation
+window.esExports = {};
+// the node.js `esm` loader won't share globals. Since this is only used
```
Standard8:
```suggestion
// The node.js `esm` loader won't share globals. Since this is only used
```

addon/content/es-modules/components/message-tags.js
siefkenj: When everything is switched to an es module, `require` will no longer be needed and this code will be used in the extension. So I guess we're left with the choice of deleting a global declaration or deleting an environment declaration. My preference would be the global declaration. Do you think the other way around?

siefkenj:
The suggestions are great :-)

addon/content/es-modules/components/message-tags.js
Standard8: No, that makes sense, lets leave it as-is.

Standard8:
Thank you for the updates.

PR #1452, opened 2020-03-02
Set tags correctly when calling browser.messages.update

### LINKED ISSUE #1450: Tagging only applies to the selected messages in the thread pane
STR:

* Select a message thread with more than one message, preferably with messages that are in different folders.
* Open and focus an older message in the thread
* Press '1'

Expected Results

The tag is applied to the older message

Actual Results

The tag is applied to the originally selected message(s)

siefkenj:
Fixes #1450


addon/content/reducer.js
```diff
@@ -197,19 +197,25 @@ function messages(state = initialMessages, action) {
       browser.messages
         .listTags()
         .then(allTags => {
+          // browser.messages.tags works via arrays of tag id/keys only,
+          // so strip away all non-key information
+          allTags = allTags.map(t => t.key);
+          // for some reason a mix of `tag.key` and `tag.id` is used in Conversations
```
Standard8: We should probably fix that at some stage. It is highly likely I got confused.

PR #1464, opened 2020-04-04
Refactor `messages` reducer

siefkenj:
This PR builds on #1463 

The `messages` reducer is currently being used for actions that don't modify the state. This PR moves all of those actions to thunks in a new `messageActions` object.

Down the line, directly dispatching an object should be changed to dispatching an action/thunk created with an action creator. This helps avoid typos in action `types`. However, this PR does not address this. It also leaves all actions that actually changed the state alone.


addon/content/messageFooter.jsx
```diff
@@ -2,39 +2,64 @@
...
+      msgUri: this.props.msgUri,
+      shiftKey: msg.shiftKey,
+    };
+    let action = null;
+    switch (msg.type) {
+      case "draft":
+        action = messageActions.editDraft(payload);
+        break;
+      case "reply":
+        action = messageActions.msgReply(payload);
```
Standard8, line 25: Seeing as these are now all within `messageActions` could we just drop the `msg` from `msgReply` etc as that is already implied.

If we do that, we could also simplify this switch down to `action = messageActions[type](payload);` maybe?

addon/content/messageHeaderOptions.jsx
```diff
@@ -124,11 +124,44 @@ class MessageHeaderOptions extends React.PureComponent {
...
+        action = messageActions.msgArchive({ id: this.props.id });
+        break;
+      case "delete":
+        action = messageActions.msgDelete({ id: this.props.id });
+        break;
+      case "classic":
+        action = messageActions.msgOpenClassic(payload);
+        break;
+      case "source":
+        action = messageActions.msgOpenSource(payload);
```
Standard8, line 159: We might need to rename some of the types, but I think dropping `msg` here as well would also help simplify - we couldn't replace the whole switch, because of the payloads, but we could treat archive & delete as specials, and then anything else we fallback to `messageActions[type](payload)` maybe?

addon/content/reducer.js
```diff
@@ -205,69 +190,59 @@ function messages(state = initialMessages, action) {
...
             tags,
           });
         })
         .catch(console.error);
-      return state;
-    }
-    case "MSG_STAR": {
+    };
+  },
+  msgStar({ id, star }) {
```
Standard8, line 200: This should probably be `changeStar` or `setStar` or something like that if you're happy to change it (I should have named it better previously).

addon/content/messageFooter.jsx
siefkenj: That was my original thought, but I thought that that might make the code too hard to follow. It's annoying that the payload that the actions take is not uniform...My original thought was that the `ActionButton` should be a connected component from the start, but I backed off because it wasn't clear what message data it would need to be passed.

addon/content/reducer.js
siefkenj: what about `starred` or `setStarred`?

Standard8: Let's go `setStarred`

addon/content/messageFooter.jsx
Standard8: Ok, no problem, let's leave the switch as it is for now.

Could we still drop the `msg` part though?

siefkenj:
I'm not sure why the tests didn't rerun, but I've addressed your comments.

Standard8:
Sometimes the tests can be a little delayed if Travis has delays on at the time.

Issue #1461, opened 2020-04-03
Redux store should be serializable


siefkenj:
The Redux store currently contains circular references and non-serilizable objects. In particular, the payload to the `"REPLACE_CONVERSATION_DETAILS"` object includes

```
summary: {
    conversation: this,
    ...
}
```
where `this` is a reference to the `new Conversation` object. This particular object causes Thunderbird 68 (on my machine at least) to freeze when attempting to print the redux store to the console.

Additionally, `neckoUrl` appears not to be searilizable (is this a special thunderbird-native wrapped object?).

Not having a serilazable store means that a lot of Redux tools can't be used and in particular, it causes an error when using Redux Toolkit to introduce the redux-thunk middleware. It also seems that the `Conversation` does a lot of storage of its own that probably should be done directly by the Redux store instead.

Getting this all sorted out should also make testing a lot easier (and possible!). However, I do not really know what the `Conversation` object is for/does nor do I know what `necoUrl` is. An explanation would be appreciated :-)

Standard8:
The `Conversation` object is currently the piece that I would say interfaces the back-end to the actual message display. Before the react-rewrite it used to drive the display and alter the DOM directly.

It currently acts as a kind of "gatherer" - gathering the conversation data and forwarding it to the message stub as needed. It also listens for updates and things like that.

We could probably start to think about if it is possible for this to be entirely driven from the stub page. I'm not quite sure yet, it'd need thinking about a bit, as the stub isn't always persistent, but that probably doesn't matter.

In the short term, I think that the `state.conversation` references in `reducer.js` could be replaced by `Conversations.currentConversation`, the `conversation` parameter can then be removed - this was probably a hang over from when I first started setting this up (or I just did it wrong for some reason).

The `neckoUrl` is slightly different to the `msgUri` that references messages, and lets us be able to stream the message. I forget what the exact differences are. In Thunderbird terms `neckoUrl` is currently a `nsIURI` object, but I don't think we need it to be - we can (and do) get the string form of it via `.spec`.

I think we can:

* In message.js in `toReactData`, change the line to get the .spec: `neckoUrl: msgHdrToNeckoURL(this._msgHdr).spec,`
* In messageIFrame.js, drop the `.spec` references.
* In the `MSG_STREAM_MSG` action we should probably also actually use the `neckoUrl` and pass it through to the `streamMessage` call and save having to get it again.

siefkenj:
[...]

PR #1465, opened 2020-04-12
Move stub initialisation to be driven by actions, refactor OS set up.

Standard8:
This starts to move the stub.xhtml initialisation to be driven via actions as discussed in #1461. At the moment, this is mainly about setting up if we're in a standalone tab or not, then for the standalone tab kicking off the conversation process.

I'm stopping there for now, as I think it is worth spending a bit more time on `Conversation` itself, and seeing if we can figure out how to get stub.xhtml driving `Conversation` in both standalone and message pane modes. That would likely make the subsequent processes easier.

However, as a bonus, I've started moving some of the back-end code to front-end. In this case the OS settings which were obtained via the back-end are now set in the front-end whilst initialising, using the browser.runtime API. That removes a couple more dependencies on misc.js.

@siefkenj I'd really value any comments you have on this.


addon/content/reducer.js
```diff
@@ -107,7 +128,167 @@ const attachmentActions = {
   },
 };
 
+// TODO: Once the WebExtension parts work themselves out a bit more,
+// determine if this is worth sharing via a shared module with the background
+// scripts, or if it doesn't need it.
+const kScrollUnreadOrLast = 0;
+const kScrollSelected = 1;
```
siefkenj, line 135: Are these supposed to be Booleans or integers?

addon/content/reducer.js
```diff
@@ -107,7 +128,167 @@ const attachmentActions = {
...
+    scrollMode == kScrollSelected &&
+    !topMail3Pane(window).gFolderDisplay.selectedMessage
+  ) {
+    scrollMode = kScrollUnreadOrLast;
+  }
+
+  if (window.frameElement) {
+    window.frameElement.setAttribute("tooltip", "aHTMLTooltip");
+  }
+  // let willExpand = parseInt(params.get("willExpand"));
```
siefkenj, line 156: Spurious comment?

addon/content/reducer.js
```diff
@@ -107,7 +128,167 @@ const attachmentActions = {
...
+          for (const id of msgIds) {
+            browser.messages.update(id, { read: true }).catch(console.error);
+          }
+        }, (await browser.conversations.getCorePref(
+          "mailnews.mark_message_read.delay.interval"
+        )) *
+          (await browser.conversations.getCorePref(
+            "mailnews.mark_message_read.delay"
+          )) *
+          1000);
```
siefkenj, line 224: This calculation is confusing inline.

addon/content/reducer.js
```diff
@@ -107,7 +128,167 @@ const attachmentActions = {
...
+      // Do this now so as to not defeat the whole expand/collapse
+      // logic.
+      if (
+        await browser.conversations.getCorePref(
+          "mailnews.mark_message_read.auto"
+        )
+      ) {
+        setTimeout(function() {
+          for (const id of msgIds) {
+            browser.messages.update(id, { read: true }).catch(console.error);
```
siefkenj, line 216: It may be overkill, but should we check *again* that `mark_message_read.auto` is true in case they changed the setting while the timeout was still waiting?

addon/content/reducer.js
```diff
@@ -107,7 +128,167 @@ const attachmentActions = {
...
+        browser.runtime.getPlatformInfo().then(info => {
+          dispatch({
+            type: "SET_OS",
+            OS: info.os,
+          });
+        }),
+        dispatch({
+          type: "SET_IN_TAB",
+          isInTab,
+        }),
```
siefkenj, line 248: Dispatching an action object directly is a blocking operation, no need for a `Promise.all` here.

addon/content/reducer.js
```diff
@@ -107,7 +128,167 @@ const attachmentActions = {
...
+        function checkStarted() {
+          let mainWindow = topMail3Pane(window);
+          if (
+            mainWindow.Conversations &&
+            mainWindow.Conversations.monkeyPatch &&
+            mainWindow.Conversations.monkeyPatch.finishedStartup
+          ) {
+            resolve();
+          } else {
+            setTimeout(checkStarted, 100);
```
siefkenj, line 265: Do we want to add a loop counter here, or trigger a warning if things are taking a really long time?

siefkenj:
I don't yet have the big picture, so I could only give minor comments. Can you explain what's challenging about tabs that require so much special treatment?

Standard8:
> I don't yet have the big picture, so I could only give minor comments. Can you explain what's challenging about tabs that require so much special treatment?

Thinking about it, tabs probably aren't actually the challenging case, the message pane is probably the more challenging currently.

Tabs get the message ids given to them in the parameters. From that they form the `Conversation` which then gets all the details and dispatches the info back.

For the message pane, we don't have those passed as parameters, but monkeypatch.js (via `window.gSummaryFrameManager.loadAndCallback.loadAndCallback`) creates the `Conversation` itself with the user selected messages.

The solution there is probably to create some sort of API that allows us to hook into `window.gSummaryFrameManager.loadAndCallback` and have some sort of callback/listener mechanism so that once the stub has loaded, it can get the messages to display, and then it creates its own `Conversation` object - for now at least. This is something I do need to think a bit more about as we need to ensure that we don't get weird race conditions with timing.

For a picture of the `Conversation` object itself, this is roughly what it currently does (in order):

* Runs a gloda (global database) query based on the initial set of messages, to hopefully get at least one gloda based message (this is one that has been indexed by the database already, if we can't we fallback to the more native style of accessing Thunderbird messages)
* Gets the collection of messages associated with that message
* Creates a `Message` object for each of the messages.
* De-duplicates & sorts
* Converts the `Message` objects to the data format we pass to the actions
* Works out which ones should be expanded or not
* Passes them to the dispatcher.

It also handles listening for updates to messages, e.g. new ones added to threads.

Some of the above, e.g. the queries, can be moved into experiment APIs. A lot of the structure we'll probably want to move into actions.

My thinking here, is if we can start off by moving creation and management of the `Conversation` object into stub, then it'll be easier to start work on re-arranging items out of the `Conversation` object and into actions / appropriate handlers. Obviously with tab handling already present in stub, that's slightly easier, hence addressing it here. The next follow-up would be to sort out the `Conversation` creation in the message pane case.

Some of the things like `setupConversationInTab` are just moved across from stub.js for convenience and so that we're starting to reduce down what is in stub.js itself. I'm expecting `setupConversationInTab` will probably get re-written and split up into actions. However for now it is quite useful as it follows a similar format to the monkeypatch code, so it'll probably be easier to combine the two in the next stage.

addon/content/reducer.js
Standard8: Integers. I've never been quite happy with these though, hopefully one day we can figure out how to simplify the scroll mode handling.

addon/content/reducer.js
Standard8: This actually ties up with code further down that's commented out. I'll move it closer.

addon/content/reducer.js
Standard8: I think it is one of those things that I'd consider highly unlikely, and if they have changed the setting just after selecting a message, then they should be expecting it to get marked as read anyway.

addon/content/reducer.js
Standard8: Yeah we shouldn't hit it, but I've added a trap just in case. I went for 10 seconds as startup could be a case where we get delayed a bit.

PR #1466, opened 2020-04-13
Remove some more msgHdrUtils deps, start cleaning up logging

Standard8:
This is a follow-on to #1465. It starts to replace some more msgHdrUtils uses, and also starts to simplify error logging which also provides more integrated output.


siefkenj:
Should I be working off this branch, or is a lot still going to change?

Standard8:
@siefkenj Not much more is going to happen on this branch. I was waiting to see if you had any more comments on #1465 before merging both of these.

The next thing I'm likely to think about is seeing if we can make the Conversation object be always created by the stub page, rather than in just the in-tab case. I think this will start shaping parts of the code that will manage & obtain the current selection.

siefkenj:
Things are looking good from my perspective. I'd be happy to hack more once these are merged.

PR #1472, opened 2020-04-27
Separated out utility functions from `contact.js` and added tests

siefkenj:
This PR starts splitting code out of `contact.js` that doesn't need to be privileged.


siefkenj:
I'm not actually sure if the emoji behaviour is correct, but I wrote the test based on how the code currently operates.

PR #1474, opened 2020-04-28
Migrate `ContactHelpers` to experiment-api

siefkenj:
Start migrating `contacts.js` to the experiment-api so we can get rid of jsm imports in `reducer.js`

### DIFF
--- a/addon/content/modules/browserSim.js
+++ b/addon/content/modules/browserSim.js
@@ -18,6 +18,9 @@ const SUPPORTED_BASE_APIS = [
   "messages",
   "runtime",
   "tabs",
+  "compose",
+  "contacts",
+  "addressBooks",
 ];
 
 const ADDON_ID = "gconversation@xulforum.org";
--- a/addon/content/modules/contact.js
+++ b/addon/content/modules/contact.js
@@ -28,20 +28,6 @@ var Contacts = {
 const defaultPhotoURI =
   "chrome://messenger/skin/addressbook/icons/contact-generic.png";
 
-/*
- * Searches a given email address in all identities and returns the corresponding identity.
- * @param {String} anEmailAddress Email address to be searched in the identities
- * @returns {{Boolean} isDefault, {{nsIMsgIdentity} identity} if found, otherwise undefined
- */
-async function getIdentityForEmail(email) {
-  const identities = await browser.convContacts
-    .getIdentities({ includeNntpIdentities: true })
-    .catch(console.error);
-  return identities.find(
-    ident => ident.identity.email.toLowerCase() == email.toLowerCase()
-  );
-}
-
 function ContactManager() {
   this._cache = {};
   this._colorCache = {};
@@ -49,7 +35,7 @@ function ContactManager() {
 }
 
 ContactManager.prototype = {
-  getContactFromNameAndEmail(name, email, position) {
+  getContactFromNameAndEmail(name, email) {
     // [name] and [email] are from the message header
     let self = this;
     email = (email + "").toLowerCase();
@@ -68,13 +54,7 @@ ContactManager.prototype = {
       return this._cache[key];
     }
 
-    let contact = new ContactFromAB(
-      this,
-      name,
-      email,
-      position,
-      this._colorCache[email]
-    );
+    let contact = new ContactFromAB(this, name, email, this._colorCache[email]);
     // Only cache contacts which are in the address book. This avoids weird
     //  phenomena such as a bug tracker sending emails with different names
     //  but with the same email address, resulting in people all sharing the
@@ -92,7 +72,7 @@ ContactManager.prototype = {
   },
 };
 
-function ContactFromAB(manager, name, email, /* unused */ position, color) {
+function ContactFromAB(manager, name, email, color) {
   this.emails = [];
   this.color = color || freshColor(email);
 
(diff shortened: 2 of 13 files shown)


addon/content/modules/contact.js
```diff
@@ -36,48 +24,20 @@ var Contacts = {
...
-    });
-  },
-};
+/*
+ * Searches a given email address in all identities and returns the corresponding identity.
+ * @param {String} anEmailAddress Email address to be searched in the identities
+ * @returns {{Boolean} isDefault, {{nsIMsgIdentity} identity} if found, otherwise undefined
+ */
+async function getIdentityForEmail(email) {
+  const browser = BrowserSim.getBrowser();
```
Standard8, line 33: Can we convert these to use the [defineLazyGetter](https://github.com/protz/thunderbird-conversations/blob/d7c365fa530f8e18a388da6ec9fdfacad08cc3be/addon/content/modules/message.js#L40-L42) in this file?

I meant to do it the other day and forgot, but should keep on top of performance.

addon/content/reducer.js
```diff
@@ -658,6 +657,25 @@ function messages(state = initialMessages, action) {
...
+          name,
+          email,
+        })
+        .catch(console.error);
+    };
+  },
+  sendEmail({ name, email }) {
+    return async () => {
+      const dest = await browser.convContacts.makeMimeAddress({ name, email });
+      await browser.compose.beginNew({ to: [dest] }).catch(console.error);
```
Standard8, line 674: Unfortunately this won't quite work right for us. We still need the information about the folder and the associated identity - otherwise if you're on a different account you'll get the wrong from address when you select select.

Whilst we can add the [identity information with TB 76 and later](https://thunderbird-webextensions.readthedocs.io/en/latest/compose.html#composedetails), we can't for TB 68.

Therefore I think we need to reimplement the current code as an API for now.

addon/content/modules/contact.js
siefkenj: I don't understand... `BrowserSim` is already defined as a lazy getter... Do you mean to define `browser` in a module and then import `browser` via a lazy getter?

siefkenj: Ah, I see your code link and understand.

siefkenj: Eventually, this will all happen in a webextension environment where `defineLazyGetter` is unavailable. Is it really worth it to do this? It's already happening in an async function.

Standard8: Once we're in the WebExtension environment, we'll already have `browser` so we won't need to be getting it manually, so if anything having it defined globally/lazily is better.

addon/content/modules/contact.js
```diff
@@ -182,20 +127,44 @@ ContactFromAB.prototype = {
...
+    const allIdentities = await browser.convContacts
+      .getIdentities({ includeNntpIdentities: true })
+      .catch(console.error);
+    const identities = await browser.convContacts
+      .getIdentities({ includeNntpIdentities: false })
+      .catch(console.error);
+    const identity = allIdentities.find(
+      ident => ident.identity.email.toLowerCase() == this._email.toLowerCase()
+    );
+    const multipleIdentities = identities.length > 1;
```
Standard8, line 145: I think the `includeNntpIdentities` parameteres are the wrong way round, but looking at the old code, I think we can simplify it anyway.

In the old code, `getIdentityForEmail` passes false as the parameter and so having `getIdentities().length > 1` default to true is a little weird - as it'd never be hit unless there was at least one identity for the email.

So I think we can only get the identities once, using `includeNntpIdentities: false`, and drop the multipleIdentities check, e.g. `extra = this._email`

addon/experiment-api/contactsApi.js
```diff
@@ -53,6 +100,61 @@ var convContacts = class extends ExtensionCommon.ExtensionAPI {
...
+
+          const browser = BrowserSim.getBrowser();
+          let tabmail = window.document.getElementById("tabmail");
+          tabmail.openTab("glodaList", {
+            collection: result,
+            title: browser.i18n.getMessage("involvingTabTitle", [name]),
+            background: false,
+          });
+        },
+        // For some reason, this needs to be implemented as an async function
```
Standard8, line 141: Side note: We don't have to do these as async, but the general design philosophy for the APIs is to have them as async, then if they do need changing later, extensions don't have to adopt.

addon/experiment-api/contactsApi.js
```diff
@@ -53,6 +100,61 @@ var convContacts = class extends ExtensionCommon.ExtensionAPI {
...
+
+          const result = await glodaQuery({
+            query: "NOUN_MESSAGE",
+            involvesItems: collection.items,
+          });
+
+          const browser = BrowserSim.getBrowser();
+          let tabmail = window.document.getElementById("tabmail");
+          tabmail.openTab("glodaList", {
+            collection: result,
```
Standard8, line 136: Have think I have a way of making this open faster when doing large queries, but I'll look at that once this has landed.

addon/experiment-api/contactsApi.js
```diff
@@ -53,6 +100,61 @@ var convContacts = class extends ExtensionCommon.ExtensionAPI {
             args
           );
         },
+        async composeNew(properties) {
+          const window =
+            properties.windowId !== null
+              ? windowManager.get(properties.windowId, context).window
+              : Services.wm.getMostRecentWindow("mail:3pane");
+          const { to } = properties;
+          composeMessageTo(to, window.gFolderDisplay.displayedFolder);
```
Standard8, line 109: I think I'd like to make this more similar to the new APIs where we can pass the identity into this one, but I'll look at that after this lands.

addon/content/modules/browserSim.js
```diff
@@ -18,6 +18,9 @@ const SUPPORTED_BASE_APIS = [
   "messages",
   "runtime",
   "tabs",
+  "compose",
+  "contacts",
+  "addressBooks",
```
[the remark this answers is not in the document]
Standard8: Err, for some reason I thought this was referring to the permissions in the manifest, not the API - sorry, you can ignore this comment.

addon/content/modules/contact.js
siefkenj: I don't quite understand this (nor have I used NNTP in the last 10 years....), but I'll drop the checks as per your suggestion.

addon/experiment-api/contactsApi.js
siefkenj: I put that comment there because I originally defined it as synchronous, but it didn't work. I couldn't figure out why (I wasn't getting a return value or something). I turned it to async and it worked. Still a mystery to me.

addon/experiment-api/contactsApi.js
siefkenj: So you want `to` to be an array? https://thunderbird-webextensions.readthedocs.io/en/68/compose.html  (which totally bit me when I was working with the webextension API...It would be much nicer to accept an array or a single object.)

Standard8: Oh I was looking at the latest ones (select at the bottom of the left-hand frame), and mainly thinking about the identity, but don't worry, I'll take a look when I get time. I want to see if it is also viable to start adding the WebExtension based ids to contacts (I think that might need the latest beta though).

addon/experiment-api/contactsApi.js
Standard8: Ah, well generally we do want these async (even if they're only sync), but the reason it has to be is because of the API definition in the schema, there's an async flag. We also want them async as eventually they'll probably be going across process, so async will be more efficient overall.

siefkenj: I had set `"async": false` in the schema though...In any case. I'm glad it's correct as it is. Should I remove the comment?

addon/content/modules/contact.js
Standard8: NNTP is a separate account type that Conversations doesn't support (because gloda, the global database, doesn't), so it is unlikely that we're going to be matching one of the news identities.

addon/experiment-api/contactsApi.js
Standard8: Ah yes, you could do.

PR #1478, opened 2020-05-13
Updated to ESLint 7, Prettier 2 and update other support modules.

Standard8:
This upgrades our modules to the latest versions, and picks up the new ESLint and prettier major updates.

@siefkenj is this going to affect anything you're working on at the moment? I can hold off on landing it if that's the case.


Standard8:
Okay, it might be a bit longer before this is ready. The Travis build shows an ESLint failure that I missed, but when I fix that, the jest run fails.

It looks like if all I do is to regenerate our package-lock.json, then the tests fail :-( something broke in a sub-dependency.

siefkenj:
This won't interfere with my work, so it seems like a good time for an upgrade! But it looks like it's the linting tests that fail, not the jest tests...?

Standard8:
@siefkenj I've just had another look, but I'm guessing you're probably going to spot this faster than I will.

To reproduce the jest failures I'm seeing, on latest master you can do:

```
$ rm -r node_modules package-lock.json
$ npm install
$ npm run test:jest
```

We then get errors such as:

```
  ● Test utility functions › getInitials extracts initials from names

    TypeError: getInitials is not a function

      26 | describe("Test utility functions", () => {
      27 |   test("getInitials extracts initials from names", async () => {
    > 28 |     expect(getInitials("")).toBe("??");
         |            ^
      29 |     expect(getInitials("X")).toBe("X");
      30 |     expect(getInitials("Tammy Smith")).toBe("TS");
      31 |     expect(getInitials("tammy smith")).toEqual("TS");

      at Object.<anonymous> (dist-dev-html/tests/utils.test.js:28:12)

  ● SvgIcon test › renders given a full path

    TypeError: React.createElement is not a function

      18 |   test("renders given a full path", async () => {
      19 |     const PATH = "/full/path/to/icon";
    > 20 |     const wrapper = enzyme.mount( /*#__PURE__*/React.createElement(SvgIcon, {
         |                                                      ^
      21 |       fullPath: PATH
      22 |     }));
      23 |     expect(wrapper.find("use").html()).toMatch(PATH);

      at Object.<anonymous> (dist-dev-html/tests/svgIcon.test.js:20:54)
```

Which look like the importer isn't working, but I can't figure out what's up.

Looks like there's a few issues against the `esm` module that could be related: https://github.com/standard-things/esm/pulls

siefkenj:
I can try to look at this over the weekend. Does it work if you revert to Jest 25? Is there a good reason to upgrade if this is the case?

Standard8:
> I can try to look at this over the weekend. Does it work if you revert to Jest 25? Is there a good reason to upgrade if this is the case?

This is without any upgrades, you can reproduce it just removing package-lock.json and letting the submodules update.

siefkenj:
So, to clarify: master is spontaneously broken if `package-lock.json` is removed?

Standard8:
> So, to clarify: master is spontaneously broken if `package-lock.json` is removed?

Not quite. On master do this:

```
$ rm -r node_modules package-lock.json
$ npm install
$ npm run test:jest
```

PR #1480, opened 2020-05-16
Upgrade failures

Standard8:
This one is just with the sub-modules updated, and shows the jest failures.

cc @siefkenj


Standard8:
I think I've got a solution - we can load some of the files as node module files.

Standard8:
The issue: Jest has started working on ES Module support (https://github.com/facebook/jest/issues/9430, https://github.com/facebook/jest/releases/tag/v25.4.0). They don't seem to limit sub-modules to patch bumps only minor bumps. So regenerating package-lock.json is bumping us to the core modules 25.4.x rather than keeping us on 25.3.x.

Partial solution:  It gets utils.test.jsx passing, and starts to work on svgIcon.test.jsx, but that isn't getting the imports for the React code right - I suspect I'm doing the wrong thing for the import.

The utils.test.jsx passes because I've effectively made the utils.js file into a nodejs module.

I'm not quite sure that's the right way to go.

The other option here would be to lock down all the jest modules and avoid updating them until we've got a way around this all.

siefkenj:
I worked on this last night and am completely stumped...`esmImport` will import the names of exported items, but their contents is always `{}`. Running the same commands on a regular `node` session work, so it's definitely a Jest issue.

I think locking down jest to `25.3.x` is the best option for now. The only alternative I could think of would be to have `babel` compile things for the `Jest` tests. This might actually be a good option as we could remove some compatibility code that way.

Standard8:
I think the error is definitely in the jest upgrade - if I update just that (with `npm install --only=dev --save-exact jest@26.0.1`) then I can see the issue.

For now, I've updated the other outdated modules in the same way, and I'll land them in #1478.

PR #1497, opened 2020-06-13
Update messages.json

Klofutar:
Update to the Slovenian translation.


Standard8:
Thank you.

Standard8:
@Klofutar FYI, I've just created issue #1553 which you can subscribe to, and I'll comment when there are new/modified string changes.

PR #1498, opened 2020-06-13
Update Ukrainian translation

a-polivanchuk:


Standard8:
Thank you for the update.

PR #1509, opened 2020-07-04
Update and fix Croatian translation

anotherus3r:
A fix and update of the Croatian translation of the add-on.


Standard8:
@anotherus3r FYI, I've just created issue #1553 which you can subscribe to, and I'll comment when there are new/modified string changes.

PR #1518, opened 2020-08-03
Updating Russian localization

Vladislav-Lyuminarskiy:
Replacing some of non-localised and Google Translate's translated strings by strings with proper Russian localization.


Standard8:
@Vladislav-Lyuminarskiy FYI, I've just created issue #1553 which you can subscribe to, and I'll comment when there are new/modified string changes.

PR #1535, opened 2020-09-21
WIP - Get Enigmail working for 78

Standard8:
This mostly works, but for some reason the subject of the message doesn't get decoded.

WIP for #1529


Standard8:
I've now got this working and will push straight to master.

PR #1544, opened 2020-10-23
Small improvement for Ukrainian translation

a-polivanchuk:


Standard8 (approved):
Thank you.

Standard8:
@a-polivanchuk FYI, I've just created issue #1553 which you can subscribe to, and I'll comment when there are new/modified string changes.

PR #1546, opened 2020-10-28
Update build.sh

jobisoft:
When I try to build you project on debian, it always fails to run build.sh, as it uses bash syntax, but identifies as sh script.


Standard8:
Thank you.

Issue #1513, opened 2020-07-18
Make the compose experiment useable and shippable


Standard8:
As referenced in #1361, there is now a [compose-experiment branch](https://github.com/thunderbird-conversations/thunderbird-conversations/tree/compose-experiment) that contains the beginnings of a compose function that is available via ctrl-shift-N (or cmd-shift-N) for writing new emails (TB78+ only).

The intention is that the widgets that are developed for it will be used in the quick reply. However, we need to get it into a shippable state before we can think about integrating it with quick reply.

This is a list of things that need to be done to get it into the shippable state, at least as a 'minimum viable product' for the compose new message function:

- [x] Styling for in a tab/window. It should look similar/the same as the function we had in TB 60, at least to begin with. Ideally it should be able to have "quick reply" mode styling as well.
- [x] Properly handle multi-line on all platforms (I think Windows will need native line endings in the message body).
- [x] Select default identity based on the selected folder (if there is one, else fallback to the existing one).
- [ ] Autocomplete for the to field.
- [x] Protection to prompt when closing the tab/window if there's been edits.
- [x] Disabling fields whilst sending.
- [ ] Send progress feedback (including error handling).

I think that's the absolute minimum I'd be comfortable with. There's more we'd want to include for shipping it as part of quick reply, but this group would be a good start.

If anyone wants to help with parts of this, please comment and let me know. If you're not quite sure what to do, I'm happy to give advance (we can also handle some of these out to separate issues if appropriate).

Standard8:
Small update, I've now updated the branch with a couple of minor improvements for the line handling - for which I've also checked it on Windows and it seems fine.

I've also done quite a bit of work on handling setting the default identity based on the selected messages, so that now should be working correctly even if you can't change the identity manually.

Standard8:
I made a bit of progress on the compose over the last couple of days, and got something working that is very basic. There's no autocomplete, no error reporting, but it can be used to send messages.

This is now on master, I might include it in the next release if I can get the error handling working a bit better.

Standard8:
Just to note, the incomplete version of this has just shipped in 3.2.7. It should be functional, though still has some rough edges.

Standard8:
This is now shipping, the only thing we haven't got is autocomplete for the to field, which I'll filed a follow-up issue for.

PR #1558, opened 2020-11-16
Switch Cu.reportError to standard JS logging (console.warn)

siefkenj:


Standard8 (approved):
Thank you for the patch, I was going to suggest console.error, but given these cases, I think we'll be ok.

Note that stub.compose-ui might go away sometime once we get quick reply going. I'm mainly keeping it around in case I need it for reference.

siefkenj:
Yeah...`stub.compose*` will need massive work to be used again. I wasn't sure if there was anything special that `Cu` did over the regular console functions.

Do you have an idea about the progress of completely moving to a web extension? Once that is done, `stub.xhtml` can import things as es modules, which would make developing UI stuff a lot easier, since it could be done in the browser with hot reloading instead of having to restart Thunderbird.

Standard8:
> Yeah...`stub.compose*` will need massive work to be used again. I wasn't sure if there was anything special that `Cu` did over the regular console functions.

I'm basically rewriting it from scratch using WebExtension APIs (and experiment ones) - see #1513. Still quite a bit to do, but I've just landed something usable on master.

> Do you have an idea about the progress of completely moving to a web extension? Once that is done, `stub.xhtml` can import things as es modules, which would make developing UI stuff a lot easier, since it could be done in the browser with hot reloading instead of having to restart Thunderbird.

Unfortunately there's quite a bit of work there still to do. Most of it I think is moving some of contacts.js, message.js & conversation.js to background scripts and some of that into the stub page. There's also likely to be some challenges about what we do wrt the iframe for displaying the actual messages.

PR #1568, opened 2020-12-13
Quoting Modifications

siefkenj:
Made a `normalizeBlockquotes` function that applies all escaping functions in order, and removed the outlook heuristic.

### DIFF
--- a/addon/content/quoting.js
+++ b/addon/content/quoting.js
@@ -7,7 +7,7 @@
 /* exported Quoting */
 
 class _Quoting {
-  canInclude(aNode) {
+  _canInclude(aNode) {
     let v =
       aNode.tagName?.toLowerCase() == "br" ||
       (aNode.nodeType == aNode.TEXT_NODE && aNode.textContent.trim() === "");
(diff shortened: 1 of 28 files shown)


addon/content/quoting.js
```diff
@@ -201,6 +201,39 @@ class _Quoting {
...
+   *
+   * @param {HTMLDocument | string} doc
+   * @returns {HTMLDocument | string}
+   * @memberof _Quoting
+   */
+  normalizeBlockquotes(doc) {
+    // We want to return the same type of object that was passed to us. We allow
+    // both a string and an HTMLDom object.
+    const origType = typeof doc;
+    if (origType === "string") {
```
Standard8, line 217: Is this set-up for later, or has it been left in?

It seems strange to have both the test and code pass in a document, but have the option for handling a string.

siefkenj: It's set up for later. I can remove it if you like.

I had a test for string input, but it failed because node does not have DOMParser in imported modules, for some reason...

Standard8: No problem, it is fine as-is, I was more curious.

Could you do a rebase/merge? github is showing conflicts.

siefkenj: Really? It says `This branch has no conflicts with the base branch` for me...

Standard8: Ah, my fault, had it on the wrong option.

Standard8 (approved):

PR #1570, opened 2020-12-13
Refactor show/hide button on blockquotes

siefkenj:
Refactor the show/hide button for blockquotes so that less of it is in the react component itself.

This PR also replaces the `div`-based button with a `button`-based button, which makes the show/hide accessible. However there's still an issue because pressing Space while the show/hide button is focused triggers a Thunderbird action in addition to toggling the blockquote (adding a stopPropogation() in the click handler didn't fix this)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -21,15 +21,18 @@ module.exports = {
       env: {
         webextensions: true,
       },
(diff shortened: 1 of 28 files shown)


Standard8 (approved):
Looks good, thank you.

PR #1575, opened 2020-12-14
Unwrap text dir Promise value

siefkenj:
`getLocaleDirection()` returns a promise. It needs to be unwrapped before being assigned to an html attribute.


Standard8 (approved):
Thank you for spotting this.

PR #1577, opened 2020-12-14
Split the reducer into several files

siefkenj:
`reducer.js` is a massive file which makes it hard to see the pieces and how they connect. This PR splits it into one file per type of action, so that they can be made "web-extension ready" one by one.


Standard8 (approved):
Thank you for this, I had started to think it was getting a bit big, but hadn't thought about it more. There's some ESLint globals/exports that can be tidied up a bit more, but I've had an idea about something that I'll try out and if it doesn't work, I'll push the changes separately.

PR #1578, opened 2020-12-26
Move to an es6 module system where possible, using webpack to package files.

Standard8:
I think I've figured out a reasonable way to make most of the stub files be modules. The basic idea is to move the import code into a separate file that's loaded as a normal javascript file (stubGlobals.js) - with defined globals. The rest of the files can then be loaded as modules, and access the globals as needed. I've made stub.js, reducer.js and reducer-deps.js into modules so far.

I'm thinking we can possibly switch to minifying all of the includes as well, which would mean that we could hopefully include react/redux/etc as normal imports as well.

Whilst this isn't 100% perfect, I think it should help with getting most of this running as modules, and making the test hacks simpler or non-existent.

@siefkenj I'm going to review your PR next (and get that landed), and then continue work on this. Hence there may be a few big changes coming. In the meantime, if you see any big issues in the idea here, let me know.


addon/content/stubGlobals.js
```diff
@@ -0,0 +1,26 @@
...
+
+/* eslint-disable no-unused-vars */
+XPCOMUtils.defineLazyModuleGetters(this, {
+  BrowserSim: "chrome://conversations/content/modules/browserSim.js",
+  Conversation: "chrome://conversations/content/modules/conversation.js",
+  topMail3Pane: "chrome://conversations/content/modules/misc.js",
+});
+/* eslint-enable no-unused-vars */
+
+window.conversationStore = {
```
siefkenj, line 17: You use `globalThis` in most places, but not here?

addon/experiment-api/msgWindowApi.js
```diff
@@ -444,7 +444,7 @@ function summarizeThreadHandler(win, id) {
     win.gMessageDisplay.singleMessageDisplay = false;
 
     win.gSummaryFrameManager.loadAndCallback(
-      "chrome://conversations/content/stub.xhtml",
+      "chrome://conversations/content/stub.html",
```
siefkenj, line 447: Yes, this is great!

siefkenj:
This looks great. We should be able to move tons of stuff to modules with this change.

Are you thinking of bundling in addition to minification? If webpack bundled stuff for us, it would remove a lot of pain points (especially ongoing troubles with Node not supporting es modules correctly. Though, maybe the newer versions of node work better?).

Standard8:
> This looks great. We should be able to move tons of stuff to modules with this change.
> 
> Are you thinking of bundling in addition to minification? If webpack bundled stuff for us, it would remove a lot of pain points (especially ongoing troubles with Node not supporting es modules correctly. Though, maybe the newer versions of node work better?).

Not sure yet, I was mainly thinking about bundling rather than minification. Though I'd like to play around a bit first and see what we need to do vs might like to do.

addon/content/stubGlobals.js
Standard8: Oops, missed that, though I'm still figuring out some of these globals. Will check this later.

Standard8:
Big update here, this switches to webpack for bundling the files for most of the various pages.

I think the add-on is working if it is built, though tests are still failing - I haven't yet worked out what is the right magic for webpack + jest + babel.

Also probably need to do some more re-arranging of some of the files, so that we can make the build script simpler (and also do things like remove redundant files/modules, and stop manually copying things).

addon/content/attachments.jsx
```diff
@@ -2,8 +2,10 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
-/* globals PropTypes, React, SvgIcon, attachmentActions */
-/* exported Attachments */
+import React from "react";
+import PropTypes from "prop-types";
+import { attachmentActions } from "./reducer-attachments.js";
+import { SvgIcon } from "./svgIcon.jsx";
```
siefkenj, line 8: Is there a particular reason you're keeping file extensions in? Since native modules cannot import `jsx`, I don't see a benefit in keeping the extension.

addon/content/es-modules/ui.js
```diff
@@ -28,55 +28,38 @@ let RTK = null;
...
-}
+React = initializedDeps.react;
+ReactDOM = initializedDeps["react-dom"];
+Redux = initializedDeps.redux;
+ReactRedux = initializedDeps["react-redux"];
+RTK = initializedDeps["redux-toolkit"];
+PropTypes = initializedDeps["prop-types"];
+useState = initializedDeps.useState;
 
 export { React, ReactDOM, Redux, ReactRedux, RTK, PropTypes, useState };
```
siefkenj, line 65: The initial idea of this file was to put all the UI exports/imports in one place so that external libraries could be imported from a single place. It looks like it's not needed anymore...? Can we just get rid of it?

siefkenj:
I haven't ever set up babel with jest directly (I use create-react-app), but maybe this is what we're looking for https://www.npmjs.com/package/babel-jest  ?  I think all the file extensions should be removed from imports, since it will likely get confusing about whether we're importing `jsx` directly or the pre-transpiled `js`.

addon/content/attachments.jsx
Standard8: Without the file extension webpack / whatever can't find the file, I think it treats it as something like a node_modules import.

```
Module not found: Error: Can't resolve './compose' in '/Users/mark/dev/thunderbird-conversations/addon/compose'
Did you mean 'compose.jsx'?
BREAKING CHANGE: The request './compose' failed to resolve only because it was resolved as fully specified
(probably because the origin is a '*.mjs' file or a '*.js' file where the package.json contains '"type": "module"').
The extension in the request is mandatory for it to be fully specified.
Add the extension to the request.
```

addon/content/es-modules/ui.js
Standard8: Most likely, still got a lot of cleaning to do :)

Standard8:
> I haven't ever set up babel with jest directly (I use create-react-app), but maybe this is what we're looking for https://www.npmjs.com/package/babel-jest ? I think all the file extensions should be removed from imports, since it will likely get confusing about whether we're importing `jsx` directly or the pre-transpiled `js`.

Thanks, I'll take a look either later or tomorrow.

addon/content/attachments.jsx
siefkenj: Very strange...I see this reference https://github.com/babel/babel/issues/8462 which says it's resolved in `@babel/runtime@7.12.0`. It seems like `babel.config.cjs` is still saying `modules: false`. Is this on purpose? Doesn't that mean things aren't bundled?

Standard8:
I think that it is all working now with the latest updates. I'll go through the diffs again tomorrow or over the weekend and then probably merge.

addon/content/attachments.jsx
Standard8: `modules: false` is for the type of output that Babel has, in this case false is keep them as es6 modules. I've currently removed that here, as now we're using Webpack, I understand that it lets Babel work it out automatically, which is better for Webpack.

PR #1579, opened 2020-12-27
Reducer cleanup

siefkenj:
Moved reducer components in `reducer-summary.js` into `summaryActions` or the reducer auomagically created by RTK.


Standard8 (approved):
Looks good, works well. Thank you.

Just to note the module patch is in-progress patch and touching pretty much all the addon/content/*.js* files. Hope to have that landed either today or tomorrow.

Issue #1553, opened 2020-11-08
Localisation Tracking Issue


Standard8:
This issue is for localisers to subscribe to, to be notified of new changes to strings ahead of them being released.

I will comment when changes land, please just subscribe to the issue. If there are any issues with the strings, please file them in new issues (hence why this issue is locked).

Standard8:
I've just pushed a couple of changes to the preferences UI - a new addition for hiding quick reply, and a reworking of the compose in a tab preference.

Difference View: https://github.com/thunderbird-conversations/thunderbird-conversations/compare/361226ae06a062fe6a180bb8a8f69d95b46fae3d..1b595a1a10e56de8ff9dd8cd1b8848ca65f42224

PR #1580, opened 2020-12-28
Update Ukrainian translation

a-polivanchuk:
Updated translation according to the latest changes.


Standard8 (approved):
Thank you. If you're not already, don't forget to subscribe to #1553 and I'll update that when there's more changes.

PR #1583, opened 2021-01-02
Merge thunks and actions into `summaryActions`

siefkenj:
* Merge thunks and actions into `summaryActions`
* Migrates `messages` actions to the RTK-based `createSlice` style.

Several of the reducers were refactors using modern Javascript.

The `msgSetIsJunk` should have the same functionality as the original code, but it seems wrong to me...Is this an unnoticed bug?

### DIFF
--- a/addon/content/attachments.jsx
+++ b/addon/content/attachments.jsx
@@ -276,7 +276,6 @@ export class Attachments extends React.PureComponent {
   showGalleryView() {
     this.props.dispatch(
       attachmentActions.showGalleryView({
-        type: "SHOW_GALLERY_VIEW",
         id: this.props.id,
       })
     );
--- a/addon/content/conversationHeader.jsx
+++ b/addon/content/conversationHeader.jsx
@@ -128,21 +128,23 @@ class _ConversationHeader extends React.PureComponent {
   }
 
   expandCollapse(event) {
-    this.props.dispatch({
-      type: "TOGGLE_CONVERSATION_EXPANDED",
-      expand: this.areSomeMessagesCollapsed,
-    });
+    this.props.dispatch(
+      messageActions.toggleConversationExpanded({
+        expand: this.areSomeMessagesCollapsed,
+      })
+    );
   }
 
   junkConversation(event) {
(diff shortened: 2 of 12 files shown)


Standard8 (commented):
Thank you for this, it is looking good.

Can you explain what you mean by this please? I couldn't quite work out what you're referring to:

> The msgSetIsJunk should have the same functionality as the original code, but it seems wrong to me...Is this an unnoticed bug?

siefkenj:
> Thank you for this, it is looking good.
> 
> Can you explain what you mean by this please? I couldn't quite work out what you're referring to:
> 
> > The msgSetIsJunk should have the same functionality as the original code, but it seems wrong to me...Is this an unnoticed bug?

The original code is

```javascript
   case "MARK_AS_JUNK": {
      // This action should only be activated when the conversation is not a
      //  conversation in a tab AND there's only one message in the conversation,
      //  i.e. the currently selected message
      browser.conversations
        .markSelectedAsJunk(action.isJunk)
        .catch(console.error);
      if (!action.isJunk) {
        // TODO: We should possibly wait until we get the notification before
        // clearing the state here.
        return modifyOnlyMsgId(state, action.id, (msg) => {
          const newMsg = { ...msg };
          newMsg.isJunk = action.isJunk;
          return newMsg;
        });
      }
      return state;
    }
```

which only changes the state if `action.isJunk == false`. Surely the state should be changed to match `action.isJunk`? Or does `markSelectedAsJunk` do modifications that get propagated and it ignores its parameter and only sets `isJunk = true`?

Standard8:
> which only changes the state if `action.isJunk == false`. Surely the state should be changed to match `action.isJunk`? Or does `markSelectedAsJunk` do modifications that get propagated and it ignores its parameter and only sets `isJunk = true`?

Ah, I think it was probably trying to optimise for the case where junk would normally get moved to the junk folder straight away. However, as that might not be always true, I'm happy for that to change here.

Standard8 (approved):
Thank you, this looks good, nice to get cleaned up a bit more.

PR #1594, opened 2021-02-02
Show/Hide Blockquote Fix

siefkenj:
Fixes race condition that occurs when detecting block quotes to show/hide.

When Conversations is deciding whether or not to put the show/hide button on a block quote, it computes the blockquote's height. This computation only works if the DOM is fully loaded. This code sets a timeout to wait for the DOM to fully load as blockquote detection runs.

A consequence of this is that the height of the `iframe` cannot be computed right away--the code needs to wait for blockquote detection to finish. This adds complication, and the solution here is probably fragile...I suggest migrating the resizing of the iframe to a well-tested library like https://github.com/davidjbradshaw/iframe-resizer-react


Standard8:
I was wondering if we might be able to do more with waiting for animation frames etc. However, that might also end up with the same issues.

I'd be happy to for us to try out the iframe resizer. Having code run by others would probably be good here.

As long as we can set all the normal iframe attributes via it, I'm fine with that.

siefkenj:
@Standard8 Were you wanting that to be done in this PR? Truth be told, I haven't used `iframe-resizer-react` before. Just read about it...

Standard8 (approved):
I think we should investigate `iframe-resizer-react`, the changes here seem a little fragile, though I don't see a better way of doing them at the moment.

However, lets go with getting this in for now, to hopefully fix the immediate issue, and then we could try that out later.

PR #1596, opened 2021-02-06
Use CSS for the contact details popup

siefkenj:
Removes the complicated logic in the `Fade` component and instead uses CSS transitions to fade the contact details in and out (fade seems to have been broken for a while anyways...).

This PR also replaces the `_getSeparator` logic with `Intl.ListFormat`. I can't quite tell, but `header.commaSeparator` and `header.andSeparator` might be able to be removed after this is merged.

I also attempted to clean up some of the `ContactDetails` object. It's still confusing...can you lay out a standard of what it's supposed to look like given what properties? If I know when it's supposed to display what, I can write some tests.

### DIFF
--- a/addon/content/messageHeader.jsx
+++ b/addon/content/messageHeader.jsx
@@ -42,7 +42,12 @@ function HoverFade({ children, popup, ...rest }) {
     document.querySelector("#popup-container") || spanRef.current;
 
   // Calculate where to render the popup
-  const pos = (spanRef.current && spanRef.current.getBoundingClientRect()) || {
+  const pos = spanRef.current?.getBoundingClientRect() || {
+    left: 0,
+    top: 0,
+    bottom: 0,
+  };
+  const parentPos = popupParentNode?.getBoundingClientRect() || {
     left: 0,
     top: 0,
     bottom: 0,
@@ -68,8 +73,12 @@ function HoverFade({ children, popup, ...rest }) {
           <div
             className={`fade-popup ${isHovering ? "hover" : ""}`}
             style={{
-              left: window.scrollX + pos.left,
-              top: window.scrollY + pos.bottom,
+              //left: window.scrollX + pos.left,
+              //top: window.scrollY + pos.bottom,
+              //left: spanRef.current?.offsetLeft,
+              //top: spanRef.current?.offsetTop + spanRef.current?.offsetHeight,
+              left: pos.left - parentPos.left,
+              top: pos.bottom - parentPos.top,
             }}
           >
             {popup}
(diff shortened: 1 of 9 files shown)


siefkenj:
I believe I fixed the issue with the overflowing popups. Everything should work now :-)

Standard8 (changes requested):
Did you use an old branch to base this on? The split options to show extra buttons at the bottom of the contacts popup seem to have come back. I got rid of that split a while ago, as I didn't think it was worth it.

The other thing I'm not quite sure about is when it gets shown. With this version of the patch if I move my mouse vertically across the contact, then the pop-up will start showing straight away.

I tried experimenting with a 400ms delay (roughly around the time of the current delay iirc), and it was better, though I was still finding that if you moved the mouse down and stopped before you get past the contact, then the contact detail was still prone to being displayed.

I'm not quite sure yet exactly how bad that is overall. Might need some more testing time.

Also I noticed that the "createFilter" text wasn't being replaced on the button.

siefkenj:
Yes, I did use an old version...(I was finishing up the work I started a year ago...). I'll update stuff to be in line with the new version and add a delay.

siefkenj:
Okay, hopefully things are good now!

addon/content/messageHeader.jsx
```diff
@@ -11,327 +12,384 @@ import { MessageTags, SpecialMessageTags } from "./messageTags.jsx";
...
+      </span>
+      {popupParentNode &&
+        ReactDOM.createPortal(
+          <div
+            className={`fade-popup ${shouldShowPopup ? "hover" : ""}`}
+            style={{
+              //left: window.scrollX + pos.left,
+              //top: window.scrollY + pos.bottom,
+              //left: spanRef.current?.offsetLeft,
+              //top: spanRef.current?.offsetTop + spanRef.current?.offsetHeight,
```
Standard8, line 102: Can we drop these lines now?

addon/content/messageHeader.jsx
```diff
@@ -11,327 +12,384 @@ import { MessageTags, SpecialMessageTags } from "./messageTags.jsx";
...
+        </span>
+      </span>
+    </HoverFade>
+  );
+}
+DetailedContactLabel.propTypes = {
+  className: PropTypes.string.isRequired,
+  contact: PropTypes.object.isRequired,
+};
+export function ContactLabel({ contact, className }) {
```
Standard8, line 176: nit: Line break before the export please

Standard8 (changes requested):
I've been trying this over the last week and not noticed any issues, so lets get this landed.

Just a minor couple of things to fix.

addon/content/messageHeader.jsx
siefkenj: Oops!

Standard8 (approved):

PR #1599, opened 2021-02-07
Devframe Updates

siefkenj:
This PR updates the dev frame to be much more featureful and on its way to being useful for development!

  * The entry point for the dev frame is now `ConversationWrapper`
  * The Locale can be dynamically set and then viewed in the devframe (sometimes you need to click between messages in order for all components within a message to update)
  * A handful of mock data has been added and `dev-frame/mock-data/threads.js` can generate several different conversation threads worth of mock data. It pulls from 100 pre-generated email objects and message objects.
  * A mock Thunderbird interface lets you choose between messages and watch them update :-).
  * All thunks are logged to the console, and Redux Dev Tools can be used to inspect any actions that get dispatched.

There was some slight gymnastics to get this all working. In particular a couple of dynamic checks about whether we're running in the browser vs. as a web extension had to be added in `messageIFrame.jsx`. As well, `browser` was changed to be imported instead of used as a global variable. However in `message-reducer.js`, the async `browser` object had to be used over the imported one in order to gain access to local storage. Hopefully this issue will become moot when conversion to a web extension is finished.

Many `browser` methods have not been shimmed, which means there are lots of errors in the console as you click around.

conversations-dev-frame


siefkenj:
@Standard8 Any idea when you'll have time to review these PRs? I have some time next week to work on things, but I think I should hold off until these changes make it in.

addon/content/es-modules/thunderbird-compat.js
```diff
@@ -56,6 +77,67 @@ if (browser.i18n) {
   browser.i18n = i18n;
 }
 
+export const Services = window.Services || {};
```
Standard8, line 80: I'm not quite sure why you're stubbing this, since `Services.locale` is only used in `getLocaleDirection` for getting the direction, and `Services.locale` is only available in privileged code.

The closest we have for WebExtensions is `browser.i18n`

addon/content/reducer-messages.js
```diff
@@ -75,7 +79,7 @@ async function setupConversationInTab(params, isInTab) {
       // teach Conversation how to handle those.
       msgUrls,
       isThreaded,
-      ++Conversations.counter,
+      ++window.Conversations.counter,
```
Standard8, line 82: Is there an advantage to using `window.Conversations` here? The fact it was in the global ESLint definition was a useful hint that it is something we need to fix.

Standard8 (commented):
Thank you for this, it looks good, just a couple of questions/comments.

(If you're not careful, you'll have soon built the whole Thunderbird UI :) )

Standard8 (changes requested):
Oops, meant to put request changes on previous comment

addon/content/es-modules/thunderbird-compat.js
siefkenj: If `Services` is not defined, there's code that errors when running in the browser. As far as I could tell, there's nothing in `i18n` that will give us the locale direction, so I figured I would stub a function that exists rather than create a new one. However, maybe a standalone `getLocaleDirection` with no deps should just be exported and used?

addon/content/reducer-messages.js
siefkenj: There was some weirdness running in the dev frame without the `window.` in front...It would complain on line 371 about `Conversations` being undefined.

https://github.com/siefkenj/thunderbird-conversations/blob/db4c9b356fdc82c00c39055ebcc385ae55dd3d4b/addon/content/reducer-messages.js#L371

I'm not sure why `Conversations` was not loaded into the global scope...It could have something to do with webpack and the way modules work? In any case, adding a `window.` in front fixed it. It will be nice when we can get rid of that global object...

addon/content/es-modules/thunderbird-compat.js
Standard8: I've just taken a look around, and I can't see the code you reference. I tried stubbing changing the additions to not use Services, and nothing else broke as far as I can tell.

Services is a privileged item, so maybe you had been trying it with one of the modules included or something?

So unless I'm missing something, a standalone `getLocaleDirection` would be fine - though I'm still debating if that's an API we should ask for in Thunderbird, or if we can just implement it ourselves.

addon/content/reducer-messages.js
Standard8: Ok, that's fine, lets keep this as you've got it for now.

addon/content/es-modules/thunderbird-compat.js
siefkenj: I just tried moving the `Services.locale` stuff into the dev frame, but I realized another issue, which is that there's currently no way to set the language from Conversations (which makes sense since it should be reading the language from Thunderbird). However, he dev frame wants to set the language and have those changes propogate through the UI.

Would you be okay with a function to change the language exported from `thunderbird-compat.js` that only gets used in the dev frame?

Standard8: That looks fine, thank you.

Standard8 (approved):

PR #1607, opened 2021-03-01
Reorganized file structure

siefkenj:
I'm not sure how you'll feel about this one, but I did a bit of reorganization of the file structure.

`.jsx` files are now in `content/components` with subdirectories for `conversation` and `message`. All redux-related files are in `reducer`, and `utils` is for pure-js utilities (that don't depend on being in a privileged context, and that can be exported as an ES module).

I also noticed the `/addon/content/i` directory appears to hardly be used? I'm also not sure whether files in `/addon/content/icons` are used...It would be nice to move `material-icons.svg` into that folder, since it seems like the right place for it. However I didn't make these changes because I'm not sure of the trickle-down consequences...


Standard8 (changes requested):
Thank you for looking at this.

I think the files in `content/i/` are unused. I suspect they may have been at one time, or it was planning or something, but in any case, looks like we can drop them.

The ones in `content/icons` are used in attachments.jsx for displaying the different attachment icons.

I'd be happy for `material-icons.svg` and `loader.gif` to both move into that directory.

For the `components` sub directory, I'm not quite so keen on that. I'd be tempted to move it to the top level, but I think with the privileged code being there that isn't quite right at the moment. So maybe leave it as you have it for now, and we'll look to tidy it up as we get more of this code separated out.

siefkenj:
By top level, do you mean directly in `/addon` instead of `/addon/content`, or do you mean something else?

You marked this PR as "Changes requested", but I'm not sure what those changes are. The removal of `/i` and moving `material-icons.svg`/`loader.gif`?

In any case, I'll wait for #1596 and #1599 to get merged, since there'll probably be some conflicts.

Another question I had was about `modules/message.js`. That file is HUGE, and it would be nice to break it apart, but if I understand correctly, it is not transpiled and `export` cannot be used in the `modules/*.js` files? That would make splitting it up more difficult.

siefkenj:
This is probably not the right thread for it, but I'm wondering what your thoughts on the process of removing `message.js` are? It looks too daunting/interconnected to be removed all at once, so I'm wondering if a parallel, unprivileged version should be created and the Redux actions should be slowly ported to depend on the underprivileged version and then when there's no more dependence, it can be removed. Of course, there may be some weird synchronization issues if message data is being read from two different objects...

One more thing: is there a way to break up the experimental api file? It seems crazy-big right now.

Standard8:
> By top level, do you mean directly in `/addon` instead of `/addon/content`, or do you mean something else?

Sorry, I meant `/addon/`.

> You marked this PR as "Changes requested", but I'm not sure what those changes are. The removal of `/i` and moving `material-icons.svg`/`loader.gif`?

Yeah, if you're happy, we might as well do those here?

> In any case, I'll wait for #1596 and #1599 to get merged, since there'll probably be some conflicts.

Oh sorry, I missed your comments there, I'll try and look at those later today.

> Another question I had was about `modules/message.js`. That file is HUGE, and it would be nice to break it apart, but if I understand correctly, it is not transpiled and `export` cannot be used in the `modules/*.js` files? That would make splitting it up more difficult.

> This is probably not the right thread for it, but I'm wondering what your thoughts on the process of removing `message.js` are? It looks too daunting/interconnected to be removed all at once, so I'm wondering if a parallel, unprivileged version should be created and the Redux actions should be slowly ported to depend on the underprivileged version and then when there's no more dependence, it can be removed. Of course, there may be some weird synchronization issues if message data is being read from two different objects...

I think there's a couple of main parts, which I should probably expand on in the discussion threads (the [roadmap briefly touched on it](https://github.com/thunderbird-conversations/thunderbird-conversations/discussions/1572) in "Move the functionality...") The short summary is that there's still a bit of functionality that handles user actions, and then there's a massive bit which handles extracting data for the message display.

We could probably look at moving some more of it to experimental APIs as a starting point which might help a bit. I'll try and remember to take a look through sometime.

> One more thing: is there a way to break up the experimental api file? It seems crazy-big right now.

We can split those files just like we've done for contacts/compose/etc. I'm a bit wary of changing that file now, since I'm hoping to be able to have a good go at getting lots of those into Thunderbird itself, or using Thunderbird APIs where we're not already. Though we could create more API files for other APIs we want to move across.

Standard8:
I'm just pushing those image removals/changes - I had a bit of spare time, so as they weren't in the other PR, I went ahead and did them.

siefkenj:
> I'm just pushing those image removals/changes - I had a bit of spare time, so as they weren't in the other PR, I went ahead and did them.

Thanks!

PR #1612, opened 2021-03-07
 Reorganized file structure (2)

siefkenj:
This PR supersedes #1607

.jsx files are now in content/components with subdirectories for conversation and message. All redux-related files are in reducer, and utils is for pure-js utilities (that don't depend on being in a privileged context, and that can be exported as an ES module).


Standard8 (approved):
Looks good, thank you for the update. I guess we can close the old PR now?

PR #1613, opened 2021-03-07
Move `material-icons.svg` to proper place in dev-frame

siefkenj:
`material-icons.svg` was recently moved to the `icons/` directory for the addon, but not for the dev-frame. This PR moves `material-icons.svg` for the dev frame.


Standard8 (approved):
Thank you, missed that, symlinks don't seem to show up in atom :(

PR #1614, opened 2021-03-14
Move the Contacts cache into `background.js`

siefkenj:
This PR moves the fetching and caching of contacts to `background.js`. It's a little janky at times, but it works...

An issue (that I think you already pointed out) is that stub.js doesn't run in a "content" process, and the `browser` object available to `stub.js` is a mocked version of the `browser` object from `background.js`. That means that `postMessage` can't be used to send messages back and forth (it's a shame that you cannot post messages to yourself...).

As a workaround, I added a `browser._background.request` method. One can register request handlers in `background.js` by adding them to the `requestHandlers` array. Every request is broadcast to all request handlers, and the first one to respond with something gets its response returned to the caller. I suspect this is similar to how it'll be in the end (when a promise-based wrapper around `postMessage`/etc. is produced).

Something I'm not too happy with is how this code currently works: It intercepts the message in `summaryActions.replaceConversation`, find the `to`/`from`/etc. fields, and repopulates them with data from the background script.

I feel like it would make more sense have a separate place in the Redux store where contact details are cached and have only the contact keys stored in the actual message object. Then, we can use a selector to combine the data as needed. This would also allow things like, if a contact is changed, its details would update live in the conversation, since there would be "one source of truth".

A question I had about this approach is: what constitutes a contact `key`? I notice the old function was called `getContactFromNameAndEmail`; it used the `email` as a key, but also used the `name` for things. Are there situations where the name returned with the message data might be different than the name returned from `browser.contacts`? Or, is the name only used when the contact is not saved... In which case, we'd want to keep around the `name` in addition to the `email` in the `message` object (unless you can think of a way to store this data separately).


siefkenj:
@Standard8 Any chance you could look at this before I incorporate your recent `contacts.js` changes?

Standard8:
> It's a little janky at times, but it works...

I wouldn't expect it to be too janky, but I guess there might be a little extra because of the extra round trip to the background process.

> As a workaround, I added a `browser._background.request` method. One can register request handlers in `background.js` by adding them to the `requestHandlers` array. Every request is broadcast to all request handlers, and the first one to respond with something gets its response returned to the caller. I suspect this is similar to how it'll be in the end (when a promise-based wrapper around `postMessage`/etc. is produced).

That seems to work reasonably well, I thought I'd had an experiment around somewhere, but I can't find it.

> I feel like it would make more sense have a separate place in the Redux store where contact details are cached and have only the contact keys stored in the actual message object. Then, we can use a selector to combine the data as needed. This would also allow things like, if a contact is changed, its details would update live in the conversation, since there would be "one source of truth".

What I'd really like to get to, or towards (and maybe not in this PR) would be something like this:

* Message selection request comes into stub.html (currently this happens via `Conversation`)
* We get the summary details for the messages via the database lookup and `browser.messages.get()` - basically just enough so that we can start displaying the message structures (e.g. have the number of msgs correct, expanded etc).
* We then fire off multiple requests at the same time, to get details for contacts, attachments, and anything else that can happen async.
* We can also fire off loading the message body for expanded messages as well
* As those requests return, update the display appropriately.

The basic idea being, get the minimum we need to work out the basic display first, and fill in the details later. We might have to be careful of flickering, but I think the layout is fixed enough that we're 

In theory, we could also do things like focus on getting what we need for the message that we're going to scroll to first, and then the other messages later. So for instance, if we're going to display three messages and scroll to the last one by default, then we could do minimal information for the first two, get all the details for the last one, and then the remaining async details for the first two.

That last theory may be over optimising - I think we should focus on making filling in the message details async, and then see what the performance / end user perspective is like.

> A question I had about this approach is: what constitutes a contact `key`? I notice the old function was called `getContactFromNameAndEmail`; it used the `email` as a key, but also used the `name` for things. Are there situations where the name returned with the message data might be different than the name returned from `browser.contacts`? Or, is the name only used when the contact is not saved... In which case, we'd want to keep around the `name` in addition to the `email` in the `message` object (unless you can think of a way to store this data separately).

`email` is the key, however, you also have to be prepared for the situation where people may have different names assigned to one email, and so you might need to display a different name with the same email address - this is why `enrichWithName` exists. `position` also has an influence on what gets displayed.


Something that might help is that [I have an old branch on doing this exact work](https://github.com/thunderbird-conversations/thunderbird-conversations/compare/8183f45...Standard8:contacts) (that I'd forgotten about when I wrote the roadmap :( ). 

I thought it still needed work to get the details right in all instances, though looking at it, it was maybe further on that I thought it was :(

Oh well. Feel free to take a look through and use some of my ideas (or not). It doesn't have to be the same and we can always move a few more things across later.

> @Standard8 Any chance you could look at this before I incorporate your recent `contacts.js` changes?

Sorry about that it was an urgent bug fix (although the contacts.js bit turned into some nice cleanup). I'll try and get some comments onto the actual changes in the next few days.

addon/content/es-modules/contact-manager.js
```diff
@@ -0,0 +1,172 @@
...
+import { freshColor } from "./utils.js";
+
+class ContactManager {
+  constructor() {
+    this._cache = new Map();
+    this._colorCache = new Map();
+    // We may ask for the same contact twice in rapid succession. In this
+    // case, we don't want to do queries multiple times. Instead we want to wait
+    // for the first query to finish. So, we keep track of all active queries.
+    this._activeFetch = new Map();
```
Standard8, line 14: nit: `_activeFetches` or `_activeQueries` maybe?

addon/content/es-modules/contact-manager.js
```diff
@@ -0,0 +1,172 @@
...
+        avatar: "chrome://messenger/skin/addressbook/icons/contact-generic.svg",
+      };
+    }
+
+    if (this._activeFetch.has(key)) {
+      // If there's an active fetch going on for this contact,
+      // caching, etc. will be taken care of by the process that spawned the
+      // fetch. Therefore, we can safely return the result of the promise directly.
+      const contact = await this._activeFetch.get(key);
+      return contact;
```
Standard8, line 46: Hmm, I think we have to be careful here. A different call to this, with the same email might have a different name with it. We still want to only do the contact fetch once, but we might not be returning the same contact object.

This is why if the cache has it, we still enrichWithName every time.

addon/content/es-modules/utils.js
```diff
@@ -2,25 +2,85 @@
...
+ * If `name` is an email address, get the part before the @.
+ * Then, capitalize the first letter of the first and last word (or the first
+ * two letters of the first word if only one exists).
  *
- * When the switch to a WebExtension is done, the actual code should be migrated
- * here and the global variable workarounds should be removed.
+ * @param {string} name
+ * @returns {string}
  */
+export function getInitials(name) {
```
Standard8, line 14: Do we want to do this function in the UI, or when we generate the contact? If we did it when generating the contact, that would potentially avoid having to do it frequently, but at the same time, it is a UI type function.

What I'm thinking about, is we could potentially merge this file into ContactManager, or maybe separate the two functions into the places where they are used, rather than having a separate module.

addon/content/reducer/reducer-summary.js
```diff
@@ -49,7 +50,10 @@ async function handleShowDetails(messages, state, dispatch, updateFn) {
 export const summaryActions = {
   replaceConversation({ summary, messages }) {
     return async (dispatch, getState) => {
-      await handleShowDetails(messages, getState(), dispatch, () => {
+      await handleShowDetails(messages, getState(), dispatch, async () => {
+        // The messages inside `msgData` don't come with filled in `to`/`from`/ect. fields.
+        // We need to fill them in ourselves.
+        await mergeContactDetails(messages.msgData);
```
Standard8, line 56: As demonstrated in #1620, I think we need to make this not await - so we fire the request and handle the response if/when we get it.

I'm happy to not address that here, as long as we're happy the architecture will support it, but I think we should aim to do it soon.

Standard8 (changes requested):
I've just commented from the architectural perspective - I think there's a couple of things that could do with improving. The comment about performance I'm happy to defer, as long as we're confident we won't need a big second rewrite to address it.

addon/content/es-modules/utils.js
siefkenj: It was originally part of `ContactManager`, I believe, but I split it out for testing. I don't think it's an expensive function, and even running it 1000 times per display probably doesn't have much of an impact.

addon/content/es-modules/contact-manager.js
siefkenj: I'm still really confused about this multiple-name-with-same-email stuff. Can you give me a concrete example of how this happens? Is it the query to `browser.contacts` that does it? If so, how does it know? Or is this information stored elsewhere?

addon/content/es-modules/contact-manager.js
```diff
@@ -0,0 +1,172 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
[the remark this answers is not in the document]
siefkenj: It certainly can be moved. My goal was to only have "entry point" files in the top-level `addon/` directory, though. If there were a `background` directory, I would think it'd make sense to relocate. What do you think?

addon/content/reducer/reducer-summary.js
siefkenj: I think if we put a per-conversation email cahce in Redux and use a reducer to create the formatted emails, then the reducer can have a nice fallback and the rest of the details will show up with they're available. This should eliminate the performance problems.

addon/content/es-modules/contact-manager.js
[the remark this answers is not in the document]
Standard8: I think I'd be fine with a `background` directory and putting all those scripts in there. Maybe in a follow-up.

addon/content/es-modules/contact-manager.js
Standard8: The multiple name/same email is generally from the perspective of the email being displayed, rather than what's in the address book, for example:

If someone has an email address example@example.com, then X might be sending to them as `foo <example@example.com>` and Y sending as `bar <example@example.com`. The names here might be subtly different, or they might be very different - but just because we displayed one first, we don't necessarily want to display this name all the time. It might be a different name is an important piece of information for the user.

However, this could apply to the situation where there's a contact stored and but without any entry in the name fields. That's likely a smaller case, but I think we should handle it regardless so that we're always displaying accurate information.

addon/content/es-modules/utils.js
Standard8: That's fine, but I think it might still be worth splitting the file, so that we're only importing the functions necessary and have a clearer distinction between background and content code.

addon/content/reducer/reducer-summary.js
Standard8: I'd be fine with that, though if we can do it per `stub.xhtml` display rather than per conversation, that might be even better - as we'd then cache emails across conversations.

Standard8, line 56: Btw, what options do we have for testing the reducers or maybe the display of contacts? I'd like to get some tests set up for the contacts display fairly soon, as I think this is quite complicated, and we should make sure we can get/keep it right.

addon/content/es-modules/contact-manager.js
```diff
@@ -0,0 +1,175 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+import { freshColor } from "./utils.js";
+
+class ContactManager {
+  constructor() {
+    this._cache = new Map();
+    this._colorCache = new Map();
```
Standard8, line 10: Just remembered, we should also listen out for changes in the address book and update these as necessary. There's some [example code here](https://github.com/thunderbird-conversations/thunderbird-conversations/compare/8183f45...Standard8:contacts#diff-cc0d0493d0507f6c06da7e674bb330d18d985f6cd6e797ab748152583ddf7fe2R15-R59).

We probably get away with it more on the current code since the contacts are cached for just the life of stub.xhtml (which changes on folder change), rather than the entire session. Though I think we should be handling it anyway.

I'm not too worried about updating the display straight away at the moment though.

We possibly should also consider a later follow-up to limit the cache sizes - I'm not sure who, but I guess some people might be seeing emails from lots of people.

addon/background.js
```diff
@@ -58,3 +74,12 @@ browser.runtime.onInstalled.addListener((details) => {
...
   }
 });
+
+// Request handler for getting contact details.
+// Accessible through browser._background.request({ type: "contactDetails", payload: contact })
+requestHandlers.push(async (msg) => {
+  if (msg.type !== "contactDetails") {
+    return null;
+  }
+  return contactManager.getContactFromNameAndEmail(msg.payload);
```
Standard8, line 84: I just had another thought - do we need to think about what happens if the conversation changes before this resolves?

I guess it would just get dropped maybe, or do we end up blocking things?

addon/content/es-modules/contact-manager.js
Standard8: I've just had a thought, not quite sure if it will work or not. Here's the idea: Contact Manager will only ever return details of the contacts themselves plus colour information. So something like:

```js
{
        emails: ["me@example.com", "me1@example.com"],
        color: "#FF0000",
        contactName: "Me", // DisplayName or FirstName + LastName (selected via PreferDisplayName)
        avatar: "chrome://messenger/skin/addressbook/icons/contact-generic.svg",
      };
```

We might need useCardName in here as well. 
The contact name is only the name from the card, not what was in the email that triggered this.

The UI then exclusively handles working out what's actually displayed for a contact.

If you want, we could separate that out to a follow-up, though I was thinking that might be easier than fixing the activeFetch issue.

siefkenj: This is close to what I was suggesting of using a selector to do the final assembly of the contact. I think it's a good idea!

addon/background.js
siefkenj: It would still get executed, but shouldn't impact anything other than the caching...

siefkenj:
@Standard8 There's be a lot of suggestions for followup PRs. Can you clarify what still needs fixing in this PR?

Standard8:
> @Standard8 There's be a lot of suggestions for followup PRs. Can you clarify what still needs fixing in this PR?

Good question! and good point.

Here's a what I think we should address for this PR:

* Put just contact-manager.js under background/, so that we don't need a second changeset to move it. The other ones can move in a follow-up.
* Move the functions out of utils.js into separate places for background/stub code. I think they can go alongside the code where they are used, and can be exported. Should still be testable this way.

Here's the follow-ups (in approximate order):

* Change ContactManager to return just the contact, and have the front-end formulate the name.
* Add tests for ContactManager and if we can for the front-end code, so that we can ensure we're ending with the right display.
* Add listening to address book updates & updating cache.
* Limit size of cache.
* Contact requests happening in async fashion alongside the rest of message loading.
* Move the other background files.

siefkenj:
I moved `background.js` and friends to a `background` folder and arranged for webpack to compile them. The only way I could get it to work was by removing the `"common"` prefix from the webpack configuration.

Standard8:
Do we need to webpack the background scripts? I'm not sure that is going to give us any advantages, since we don't need it for the jsx processing, and we can generally run the latest javascript features without needing processing.

siefkenj:
> Do we need to webpack the background scripts? I'm not sure that is going to give us any advantages, since we don't need it for the jsx processing, and we can generally run the latest javascript features without needing processing.

We don't strictly _need_ it, but it makes development much easier. Using native es-modules, I had to edit `build.sh` to manually copy the imported files. Whenever I imported something from a different location, I would often forget to do this and then have to spend time debugging...Using webpack avoids all of this. It also gives compile-time warnings if modules cannot be found, which has saved me a lot of time!

Standard8:
> We don't strictly _need_ it, but it makes development much easier. Using native es-modules, I had to edit `build.sh` to manually copy the imported files. Whenever I imported something from a different location, I would often forget to do this and then have to spend time debugging...Using webpack avoids all of this. It also gives compile-time warnings if modules cannot be found, which has saved me a lot of time!

On the flip side, it makes it harder to debug, due to the packaging that webpack does, and the variable name changing and things.

However, I think we'll try it for now and see how it goes, we can always change it back later :)

Standard8 (approved):
Ok, lets go with this. I've filed #1624 for the follow-ups I mentioned so that we don't have to refer to a closed issue.

PR #1625, opened 2021-04-14
Rework compose page to share the component with quick reply

Standard8:
This is something I've just started working on to allow the compose component to be used with quick reply as well. There's still a lot to do to get this working good enough to land and tidied up, but working towards it is a good next step for quick reply.


Standard8:
@siefkenj one thing I'm puzzled about here, is that the dev frame doesn't seem to handle the reducer actions. For instance, when I start typing in the form, `composeActions.setValue()` gets called correctly, but the call into `composeSlice.reducers.setFromDetails()` never happens.

Are we missing something in the dev frame set-up?

If it doesn't work easily it isn't a big issue, mainly wondering why I can't simulate it there.

siefkenj:
ATM, reducers aren't imported in the dev frame because they had dependencies on privileged TB code. But, maybe that's been cleaned up now?

Standard8:
> ATM, reducers aren't imported in the dev frame because they had dependencies on privileged TB code. But, maybe that's been cleaned up now?

Ah that makes sense. For the compose side we definitely shouldn't be since that can run in a WebExtension page of its own. For the others I'll need to take another look.

Can we hook up just one of the reducers?

Standard8:
> Can we hook up just one of the reducers?

I've just worked this out.

Standard8:
@siefkenj I'm thinking I might land this in the next day or so (with the quick reply disabled again) due to the restructuring. If you have any general comments on the architecture it'd be useful.

The basic idea is:

* addon/content/components/compose/ to share common files.
* addon/compose for the separate ctrl/cmd-n compose page.
* addon/content/components/quickreply to manage the quick reply wrapping (expand/collapse, init etc)
* addon/content/reducer/reducer-compose.js for the compose reducer, and there will probably two more - one for quick reply, and another one for the separate compose page.

Standard8:
(i.e. I think it is worth landing the restructuring at this stage).

addon/content/components/compose/composeWidget.jsx
```diff
@@ -0,0 +1,118 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+"use strict";
```
siefkenj, line 5: Isn't `"use strict"` automatically implied for es modules?

addon/content/components/compose/composeWidget.jsx
```diff
@@ -0,0 +1,118 @@
...
+
+import { composeActions } from "../../reducer/reducer-compose.js";
+import React from "react";
+import * as ReactRedux from "react-redux";
+import PropTypes from "prop-types";
+import { TextArea, TextBox } from "./composeFields.jsx";
```
siefkenj, line 11: Do we have a convention of camelCase vs hyphenated file names?

addon/content/components/compose/composeWidget.jsx
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  React.useEffect(() => {
+    window.addEventListener("beforeunload", checkBeforeUnload);
+    return () => {
+      window.removeEventListener("beforeunload", checkBeforeUnload);
+    };
+  });
+
+  // TODO: We may need to be able to jest before we can remove these
```
siefkenj, line 70: What does "being able to jest" mean?

addon/content/components/compose/composeWidget.jsx
```diff
@@ -0,0 +1,118 @@
...
+    html.setAttribute("os", OS);
+  }
+
+  return (
+    <div className="compose">
+      {INPUT_FIELDS.map((Item, i) => (
+        <Item.component
+          {...Item.props}
+          disabled={Item.props.disabled || composeDetails.sending}
+          key={i}
```
siefkenj, line 85: We can use `Item.props.name` instead of `i`, since it should be fixed and unique.

addon/content/components/compose/composeWidget.jsx
```diff
@@ -0,0 +1,118 @@
...
+  OS: PropTypes.string,
+  composeDetails: PropTypes.object.isRequired,
+};
+
+export const ComposeWidget = ReactRedux.connect((state) => {
+  return {
+    OS: state.summary.OS,
+    composeDetails: state.compose,
+  };
+})(_ComposeWidget);
```
siefkenj, line 118: With hooks, `connect` is unneeded.

In the actual component, you can do
```javascript
const dispatch = ReactRedux.useDispatch();
const {OS, composeDetails} = ReactRedux.useSelector((state) =>
    ({OS: state.summary.OS, composeDetails: state.compose}));
```

addon/content/components/compose/composeWidget.jsx
```diff
@@ -0,0 +1,118 @@
...
+    },
+    component: TextBox,
+  },
+  {
+    props: {
+      name: "body",
+      disabled: false,
+    },
+    component: TextArea,
+  },
```
siefkenj, line 44: Are there going to be a lot more of these? I wonder if it's worth the abstraction. Will `TextArea` and `TextBox` maintain the same props forever? If not, it seems weird to batch-create them when there are only four things.

siefkenj:
>     * addon/content/components/compose/ to share common files.
> 
>     * addon/compose for the separate ctrl/cmd-n compose page.
> 
>     * addon/content/components/quickreply to manage the quick reply wrapping (expand/collapse, init etc)
> 
>     * addon/content/reducer/reducer-compose.js for the compose reducer, and there will probably two more - one for quick reply, and another one for the separate compose page.

Initially this sounded like a lot of different places to spread files out, but the more I think about it, the more reasonable it sounds. So the basic idea is `addon/XXX` is an entry point, except for `addon/content`. And everything else is in subfolders of `addon/content`?

addon/content/components/compose/composeWidget.jsx
Standard8: I think I generally prefer camelCase. Though I've just realised we do have a mixture.

addon/content/components/compose/composeWidget.jsx
Standard8: I think I meant "be able to test"... or maybe something in the jest tests was playing up. I'll take a look.

addon/content/components/compose/composeWidget.jsx
Standard8: Thanks, I'll give that a go.

addon/content/components/compose/composeWidget.jsx
Standard8: I'm not totally sure yet. The "from" text box will probably change into a drop-down of identities, the other ones we'll probably need to add some sort of autocomplete, and a few more boxes.

So they will get more complicated, even if they aren't now.

Standard8:
> Initially this sounded like a lot of different places to spread files out, but the more I think about it, the more reasonable it sounds. So the basic idea is `addon/XXX` is an entry point, except for `addon/content`. And everything else is in subfolders of `addon/content`?

Pretty much. `addon/XXX` is the entry point for the pages, compose, assistant etc.

Eventually I'd like to drop `addon/content` and replace it by `addon/stub` or something like that, but we can't really do that until we get rid of the privileged code (there's other code in the APIs that registers the files in `addon/content` so they can be accessed). For now the `content` name also works better for the Thunderbird add-on reviewers as they'll recognise that as meaning privileged code.

Effectively `addon/content/modules` should either get merged into various parts of stub, or into background scripts. `pdfviewer` is going away as soon as we can drop 78, and the rest of that is pretty much stub related.

addon/content/components/compose/composeWidget.jsx
Standard8: I was confused when I originally wrote the response - yes I'll drop the abstraction, I don't think it is helping that much.

Standard8:
I've fiddled with this a bit more and added a basic architecture for expanding/collapsing. This also means you can send a message from within the quick reply area, although it might not have the correct from address, and it doesn't quote/use the parameters from the rest of the message.

Still, it is a useful starting point which will make it easier to play around with getting the various parts going.

Issue #1590, opened 2021-01-21
DKIM Verifier not working with Conversations in Thunderbird 78+


farfallosa:
Thunderbird version: 78.6.1 
Thunderbird Conversation version: 3.2.11
Operating System: Windows 8.1, Firefox 84.0.2 

in former versions of TC there was a information within TC mail view, wheter DKIM is used or not.
In classic view of the mail I'll find this information. It would be great to have this info in TC back!

Standard8:
You probably also have the DKIM verifier add-on installed. Due to changes in Thunderbird, the two add-ons are not currently compatible. We have started discussing it here, but still need to do the work to get it running again:

https://github.com/lieser/dkim_verifier/issues/203

farfallosa:
@Standard8 Yes, I have DKIM verifier add-on installed. But even after deactivating that add-on (and restarting TB) the DKIM Label in TC does not appear. I tried it for several times, but it seems that the bug persists even after deactivating DKIM verifier Add-on.
Yes, absolutely, I would appreciate DKIM running on TC again.

Standard8:
We've just landed the Conversations side of this fix on main branch. It'll probably be a few weeks before it is released.

There is also a DKIM side of the fix, the release period is unknown at the moment (see https://github.com/lieser/dkim_verifier/issues/203)

Issue #1431, opened 2020-01-19
Migrate add-on adapter plugins to use WebExtension APIs/messaging


Standard8:
With the recent switch to React, and the changes around how the UI works, I think we could change the adapter plugins (addon/content/modules/plugins/ to use WebExtension APIs / messaging that conversations provides for the purpose.

This would also help with a better API definition for supporting other add-ons that want to integrate with Conversations.

We'll probably want to use some type of message registration system, and have WebExtension APIs exposed for integrating into the back end of those add-ons. If they're migrating as well, then at some stage we might need to do a bit more than that.

Standard8:
+ also add docs.

Standard8:
This is now done in #1690

PR #1690, opened 2021-07-25
Allow other addons to add pills via message API

lieser:
- add listener for an `addPill` message, which allows other add-ons to add tags
- started to remove/replace DKIM plugin specific stuff

#1431 #1590

### DIFF
--- a/addon/background/window.js
+++ b/addon/background/window.js
@@ -56,14 +56,37 @@ export class Window {
 
     browser.convMsgWindow.onSummarizeThread.addListener(async () => {});
 
+    /**
+     * @typedef {"normal"|"success"|"warning"|"error"} Severity
+     */
+    /**
+     * @typedef {object} AddPillMessage
+     * @property {"addPill"} type
+     *  The type of the received message.
+     * @property {number} msgId
+     *   The id of the associated Message from the WebExtension APIs.
+     * @property {string?} icon
+     *   The optional icon of the pill.
+     * @property {string} message
+     *   The text of the pill.
+     * @property {string[]} tooltip
+     *   The tooltip of the pill.
+     * @property {Severity} severity
+     *   The severity of the pill.
+     */
     browser.runtime.onConnectExternal.addListener(async (port) => {
       port.onMessage.addListener((msg) => {
         if (msg.type != "addPill") {
           return;
         }
+        /** @type {AddPillMessage} */
+        const pillMessage = msg;
         browser.convMsgWindow.addSpecialTag({
-          msgId: msg.msgId,
-          message: msg.message,
+          msgId: pillMessage.msgId,
+          icon: pillMessage.icon ?? "material-icons.svg#edit",
+          classNames: pillMessage.severity,
+          message: pillMessage.message,
+          tooltip: pillMessage.tooltip,
         });
       });
     });
--- a/addon/background/window.js
+++ b/addon/background/window.js
@@ -65,14 +65,14 @@ export class Window {
      *  The type of the received message.
      * @property {number} msgId
      *   The id of the associated Message from the WebExtension APIs.
-     * @property {string?} icon
-     *   The optional icon of the pill.
+     * @property {Severity} [severity]
+     *   The severity of the pill. Defaults to normal.
+     * @property {string} [icon]
+     *   The optional icon of the pill. Musst be an "moz-extension://" url.
      * @property {string} message
      *   The text of the pill.
-     * @property {string[]} tooltip
-     *   The tooltip of the pill.
-     * @property {Severity} severity
-     *   The severity of the pill.
+     * @property {string[]} [tooltip]
+     *   The optional tooltip of the pill.
      */
     browser.runtime.onConnectExternal.addListener(async (port) => {
       port.onMessage.addListener((msg) => {
@@ -81,12 +81,20 @@ export class Window {
         }
         /** @type {AddPillMessage} */
         const pillMessage = msg;
+
+        if (
+          pillMessage.icon &&
+          !pillMessage.icon.startsWith("moz-extension://")
+        ) {
+          pillMessage.icon = undefined;
+        }
+
         browser.convMsgWindow.addSpecialTag({
           id: pillMessage.msgId,
+          classNames: pillMessage.severity ?? "normal",
           icon: pillMessage.icon ?? "material-icons.svg#edit",
-          classNames: pillMessage.severity,
           message: pillMessage.message,
-          tooltip: pillMessage.tooltip,
+          tooltip: pillMessage.tooltip ?? [],
         });
       });
     });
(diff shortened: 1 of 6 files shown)


addon/background/window.js
```diff
@@ -55,6 +55,41 @@ export class Window {
...
+     * @typedef {"normal"|"success"|"warning"|"error"} Severity
+     */
+    /**
+     * @typedef {object} AddPillMessage
+     * @property {"addPill"} type
+     *  The type of the received message.
+     * @property {number} msgId
+     *   The id of the associated Message from the WebExtension APIs.
+     * @property {string?} icon
+     *   The optional icon of the pill.
```
Standard8, line 69: Should probably note that this needs to be a full moz-extension:// url. Would be good if we could add a check for that in the listener as well - we could fallback to the internal icon if it is an incorrect url.

Standard8 (commented):
I've not tried it out yet, but this is looking good. Just one comment in-line.

addon/background/window.js
lieser: Ok, will add a check for it.

Another think I forgot to mention is that we need to decide on the handling of the optional properties.
Currently they must be explicitly set (at least according to this typedef), i.e. the `icon` must be set to `null` and the `tooltip` to an empty array. Don't know if you also want to allow `undefined` for both, i.e. so that a user can just leave them out in the message.
And we may want to make the `severity` optional too.

Standard8: Yeah, I think we can make icon, tooltip and severity optional.

lieser: Added the check for `moz-extension://`, and made all besides `msgId` and `message` optional.

addon/content/conversation.css
```diff
@@ -772,28 +768,37 @@ html[dir="rtl"] .attachmentsDropDown .arrow {
...
 
-.dkim-signed {
+.special-tag-ext-icon {
+  margin-inline-end: 0.2em;
+}
+
+.messageHeader img {
+  /* Override ".messageHeader img" to default ".icon" */
+  vertical-align: -0.4em;
+}
```
lieser, line 791: A little dirty workaround, as I do not know to which other elements `.messageHeader img` is applied, and if the `vertical-align: text-top` is wanted there.

lieser:
> `messagesSlice.reducers.msgAddSpecialTag` will only add if they're a different type, but in this case the type is undefined as it isn't specified in window.js.

Don't know exactly what you mean with this, but I think I found the problem. Unfortunately I currently don't see an easy fix I could do, maybe you have an idea.

The problem is the state if `reducers.msgAddSpecialTag` is called. It can happen that `state.msgData` does not yet contain any data (for first message viewed after start), or that it still contains the last viewed messages, instead of the new ones, for which `browser.messageDisplay.onMessagesDisplayed` was already called.

In both cases `modifyOnlyMsg()` does not see the message it should add the tag for, and the tag is lost.

> `msgAddSpecialTag` is sort of wrong anyway, as it should allow updating if the type is the same.
>
> Though I'm also thinking that maybe we should allow multiple types from other add-ons, maybe using some sort of id field or something?

Again, unsure what exactly you mean with `type` here.

Adding multiple pills/tags should be possible. At least if I send the (same) `AddPillMessage` two times to conversation, too pills show up.

And agree, updating an existing pill could be something other add-ons may want/need. Should be achievable by using a combination of an id included in `AddPillMessage` message, and the sender id of the message.

Btw, do you want to keep the *pill* naming for the interface for other addons? Internally, conversation seems to rather use *tag* for it.

lieser:
One possibility to avoid the current timing issue would be that conversation forwards the `browser.messageDisplay.onMessagesDisplayed` (with and additional event ID) through the established connection. And add-ons must react to that, instead of `browser.messageDisplay.onMessagesDisplayed`.
This would allow Conversation to ensure that the state is already updated, and to track if an incoming `AddPillMessage` (which would need to include the event ID) is for the currently viewed messages. But it would mean a more complicated integration between the addons.

Still hoping that you have a simpler idea what Conversation could do to avoid this.

Standard8:
Thanks for the comments. I've just pushed a change that should fix the timing of `onMessagesDisplayed` and make it happen for conversations in tabs/windows. That was a regression from some rework I did recently - most of the message processing is now happening within the stub.xhtml page, but the completion was still being indicated from the module that was finding the messages.

I think it should be better now. I'm not quite sure if there could be a timing issue if messages threads are switched whilst loading is happening, but we'll see how it goes.

It also let me simplify some code, which is always good :)

> Btw, do you want to keep the _pill_ naming for the interface for other addons? Internally, conversation seems to rather use _tag_ for it.

Yeah, not sure. I think we could probably change that to tag. Possibly a little confusing wrt message tags, but might as well be consistent for now.

lieser:
The timing fix seems to work, in my tests the added pills did now always show up.

Unfortunately there is still an issue with the "show this conversation in a new tab" and `onMessagesDisplayed`. For me the array of messages is always empty.

If a thread is viewed, `onMessageDisplayed` now gives `null` for the message. That is completely fine for me (as I am ignoring that event anyways for the conversation view), but if possible I think it would be better if it is not called at all.
It could result in potential problems for other addons, wo assume that the message always has a proper value if `onMessageDisplayed` is called. 

> Yeah, not sure. I think we could probably change that to tag. Possibly a little confusing wrt message tags, but might as well be consistent for now.

Completely up to you. As there is only one function to call, the naming will be consistent for other addons regardless of that the name is.
I would say it depends on if you have plans to more clearly distinguish in Conversationmake between the normal Thunderbird tags and other tags (i.e. pills), If you do, I would keep the pills name. If you do not plan to distinguish them, I would change it to tags.

Standard8:
> The timing fix seems to work, in my tests the added pills did now always show up.

Great!

> If a thread is viewed, `onMessageDisplayed` now gives `null` for the message. That is completely fine for me (as I am ignoring that event anyways for the conversation view), but if possible I think it would be better if it is not called at all.
> It could result in potential problems for other addons, wo assume that the message always has a proper value if `onMessageDisplayed` is called.

Oh yeah, that's wrong, I think what I'll do for now is take that notification out for the new tab. I need to do some work there to see if I can somehow override TB's extension API code. 

> > Yeah, not sure. I think we could probably change that to tag. Possibly a little confusing wrt message tags, but might as well be consistent for now.
> 
> Completely up to you. As there is only one function to call, the naming will be consistent for other addons regardless of that the name is.
> I would say it depends on if you have plans to more clearly distinguish in Conversationmake between the normal Thunderbird tags and other tags (i.e. pills), If you do, I would keep the pills name. If you do not plan to distinguish them, I would change it to tags.

Ok, lets keep it as pill for now :)

addon/content/conversation.css
Standard8: This doesn't actually appear to be used, since the pill is in `.messageBody`. I'm going to merge for now, remove both of the `.messageHeader img` options (there's no `img`s in `.messageHeader`), and we can always tweak in a follow-up.

Standard8 (approved):
Thank you for your work on this, great to have some form of support back again!

Issue #1746, opened 2021-11-03
Can't view messages in tabs when folder names contain non-ascii characters


msprotz:
Hi,

Since Thunderbird 91, I have been unable to open messages in folders that contain non-ASCII characters. If I browse to a folder that contains a non-ASCII character, then click on a message, I get:

```
Unable to get  imap-message://jonathan.protzenko%40gmail.com@imap.gmail.com/â�­ [GitHub]#42287  — returning null instead 3 misc.js:50:13
```

I am unsure whether this is a Conversations bug or a Thunderbird bug but opening such messages with the classic reader works.

Thanks,

Jonathan

Standard8:
I can reproduce this - only affects opening in a new tab/window afaict. I think it is probably a Thunderbird issue - there's something similar going on with not being able to restore message tabs in the classic reader when the messages are in an folder with non-ascii characters. That case looks like it is using the same functions.

I'll try and file a bug on this later.

msprotz:
Thanks Mark!!

Standard8:
This is definitely a Thunderbird issue. They changed how the message uris work which now include non-ascii interfaces, and didn't update all the interfaces to cope.

I'm putting up fixes to get the Conversations related parts fixed in this bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1739784

Standard8:
The patch has just been released in Thunderbird 91.3.2. This is fixed in my tests locally.

msprotz:
Wonderful. Thanks so much.

PR #1758, opened 2021-11-23
trying to fix message uri encoding..

erqan:
if folder name contains non latin character, it fails to get message from URI


Standard8:
Hi, thank you for trying to fix this, however this is already fixed if you use the latest version of Thunderbird (91.3.2) - see #1746 

This was actually an issue caused by changes within Thunderbird itself, so we had to fix them there - see the bug referenced.

PR #1811, opened 2022-03-17
Update to the Slovenian translation

Klofutar:


Standard8 (approved):
Thank you for the updates.

Issue #1453, opened 2020-03-03
Remote Content Preferences should be available 


fyodorovich:
* **Thunderbird version:**  	68.5.0
* **Thunderbird Conversation version:** 	3.0.16
* **Operating System:** Mac OS X 10.13.6

**What did you do?** Received an HTML formatted email. I wanted to view the images which were being blocked by TB.

**What did you expect to happen?** I expected to be able to find the preferences button that would allow me control the remote content settings for that message. The options to control the display of the remote content is important. We need to be able to access that functionality. 

**What actually happened?** I wasn't able to find the button until I disabled TB conversations. As soon as I did that, button was displayed.

Standard8:
Do you not see a bar like this?

<img width="716" alt="Screenshot 2020-03-04 at 08 26 42" src="https://user-images.githubusercontent.com/520844/75863564-9ae20e00-5df8-11ea-82d2-681d107b420b.png">

This was on the github email I got for this issue.

Standard8:
Sorry, but without more information I can't help here.

fyodorovich:
No, that bar is not visible. That is exactly the sort of thing I was looking for. 

( Sorry, I thought I had replied 14 days ago when you posted. )

PR #1844, opened 2022-05-24
Work-around for missing "Remote Content" bar (#1453)

DabeDotCom:
I was getting `TypeError: context is undefined` on `msgWindowApi.js:364`
(See #1453)

After poking at it a little bit, I discovered that `msgWindowApi.js` is,
for whatever reason, actually calling the `monkeyPatchAllWindows()` from
`contactsApi.js` (!) which doesn't accept/passthru a `context` parameter.

Rather than solve the hard problem of namespacing/scoping the conflicting
function definitions, however, I simply took the path of least resistance
and added the `context` parameter to `contactsApi.js`.  :-}

### DIFF
--- a/addon/experiment-api/contactsApi.js
+++ b/addon/experiment-api/contactsApi.js
@@ -489,10 +489,10 @@ function waitForWindow(win) {
   });
 }
 
-function monkeyPatchAllWindows(windowManager, callback) {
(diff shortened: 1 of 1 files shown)


Standard8 (approved):
Thank you for the fix.

I really don't like the way the scopes between these api files are shared but that's not your issue :)

PR #1914, opened 2022-10-12
Fixes #1913: Show 'reply all' option if reply-to is different from the sender

### LINKED ISSUE #1913: No 'Reply All' option for emails from a mailing list with reply-to header
* **Thunderbird version:** 102.3.1
* **Thunderbird Conversation version:** 4.0.24
* **Operating System:** Linux Fedora 36

**What did you do?**
Received an email from a mailing list with From and To being a list address and Reply-To being an original sender.

**What did you expect to happen?**
Expected to have a 'Reply All' option to send a reply to both the mailing list and the original sender.  Classic view allows that.  It's not a first option in the classic view, but it is available in a drop-down menu.

**What actually happened?**
The only options are 'Reply' and 'Reply to list'.   First replies only to the 'Reply-To' address, the second only replies to the list.  But there is no option to reply to both.

On a different mailing list that is using original sender in a From header and has no Reply-To header, the 'Reply All' option is available.
Unfortunately, lists has to move away from that method due to DMARC issues.

igsilya:
Mailing lists with 'Mangle From' configuration swaps 'From' with the mailing list address and keeping the original sender in the 'Reply-To' header.  Currently only 'reply' and 'reply to list' options are available in this case, but there should be an option to reply to both the list and the original sender.

Taking the 'reply-to' header into account while checking for multiple "recipients" to fix the problem.

### DIFF
--- a/addon/content/reducer/contacts.js
+++ b/addon/content/reducer/contacts.js
@@ -169,7 +169,7 @@ export async function mergeContactDetails(msgData) {
 function hasMultipleRecipients(message) {
   let seen = new Set();
   let count = 0;
-  for (let field of ["from", "to", "cc", "bcc"]) {
+  for (let field of ["from", "to", "cc", "bcc", "replyTo"]) {
     // TODO: The ?? and subsequent !contact currently helps some of the tests to pass.
     let contacts = (field == "from" ? [message[field]] : message[field]) ?? [];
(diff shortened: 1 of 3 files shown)


Standard8 (approved):
Thank you, this looks reasonable to do.

PR #1989, opened 2023-06-29
Use Services global variable

arai-a:
Services.jsm is planned to be removed in Firefox 117 cycle in https://bugzilla.mozilla.org/show_bug.cgi?id=1780695 . Services global variable is in all system globals from version 104 https://bugzilla.mozilla.org/show_bug.cgi?id=1667455 , and those code doesn't have to import Services.jsm if the strict_min_version is 114.

### DIFF
--- a/addon/content/stubWrapper.js
+++ b/addon/content/stubWrapper.js
@@ -4,8 +4,6 @@
 
 "use strict";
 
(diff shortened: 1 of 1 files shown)


Standard8 (approved):
Thank you!

PR #2037, opened 2023-10-21
Update messages.json [Ukrainian translation]

artpoli:


Standard8:
@artpoli The translations for these calendar buttons were taken from the official Thunderbird translations, for example, if you view an email in the classic reader, you'll see the same translations as I added here.

Hence, I'm not too keen on diverging, since this would introduce inconsistencies, and generally the L10n teams try to be consistent across the UI. 

Would you consider proposing these as updates to Thunderbird via https://pontoon.mozilla.org/uk/thunderbird/ ? If they get accepted, then I would gladly accept the updates here as well.

artpoli:
Hey @Standard8 
Thanks for clarifying! As a Ukrainian language manager on Pontoon, I'll submit these changes to Thunderbird localization and drop a link to the hg changeset here :)

artpoli:
So, I reviewed and updated the related translations in Thunderbird, and it seems that now they're consistent with this PR.
[Pontoon: Update Ukrainian (uk) localization of Thunderbird](https://hg.mozilla.org/projects/comm-l10n/rev/db4b867088a8549ce5c55c7a91bd22464f2feb81)

Standard8:
Aha, it is great that we can update and improve both. Thank you for doing both of those.

Standard8:
This probably won't go out straight away, but I'll get it in whatever is the next release after the one that's in the approval queue.

Issue #666, opened 2012-10-29
Pdf-preview problem


Fishever:
Hi there. I kinda have a problem with pdf-preview. I'm using Thunderbird to manage several Email accounts. The weird thing is that pdf-preview works fine only with my imap account (gmail) and for all the other pop3 accounts it's not working at all (like hotmail). I can only see blank page even without tab title, just blank. When I move the mail to my gmail account and it works again......The error console shows "URIError: malformed URI sequence
chrome://conversations/content/pdfviewer/wrapper.js line: 122"
I'm only an usual user and can't figure out the reason.It's not a big issue but still I hope it can be fixed if it's possible. I'm using thunderbird 16.0.1 (chinese) with conversation version 2.5.2.

PR #2080, opened 2024-02-17
Improve appearance of quick reply send button

### LINKED ISSUE #1933: 'Send' button in quick reply box appears with grey text on white background in dark theme
* **Thunderbird version:** 102.5.1
* **Thunderbird Conversation version:** 4.0.25
* **Operating System:** Windows 10

**What did you do?**
Enabled dark theme in TB.

**What actually happened?**
"Send" button in quick reply box appears with grey text on white background, making it unreadable:

send button is grey on white

**What did you expect to happen?**
"Send" button in quick reply box appears with readable text color.
If I enable colors substitute in Thunderbird settings button appears with readable text color:
send button is readable

fezjo:
Fixes #1933 

Improve appearance of send button in conversation compose box. Improvements include:
- support for dark mode
- cursor and color change on hover (old design didn't react in any way)
- more padding and boxier design more in line with modern Thunderbird style

Comparison:

Current Light
before-light

Current Dark
before-dark

New Light
after-light-normal
after-light-hover

New Dark
after-dark-normal
after-dark-hover

It seems like the surrounding box doesn't resize properly. The bottom edge of the button is close to the border and _maybe_ on some other machines it will stick out. I am not really well versed in HTML and CSS so I don't know how to fix the layout properly. However, this change as it is is still an improvement.

Future improvements should also include changing links `discard`, `Forward this conversation`, `Print this conversation` to button style. It seems like the UI has been neglected and can be easily improved.

### DIFF
--- a/addon/content/components/compose/composeWidget.css
+++ b/addon/content/components/compose/composeWidget.css
@@ -73,7 +73,6 @@ input:focus {
   padding: 0.3em 1em;
   background-color: -moz-dialog;
   color: -moz-dialogtext;
-  cursor: pointer;
   border: 1px solid ThreeDShadow;
   border-radius: 3px;
 }
--- a/addon/content/components/compose/composeWidget.css
+++ b/addon/content/components/compose/composeWidget.css
@@ -68,12 +68,17 @@ input:focus {
 }
 
 .compose > .buttons > button {
-  padding: 0 5px;
-  margin-inline-end: 3px;
-  background-color: white;
+  padding: 0.3em 1em;
+  margin-inline-end: 0.2em;
+  background-color: -moz-dialog;
+  color: -moz-dialogtext;
+  cursor: pointer;
   border: 1px solid ThreeDShadow;
-  height: 1.86em;
-  border-radius: 7px;
+  border-radius: 3px;
+}
+
+.compose > .buttons > button:hover {
+  filter: brightness(85%);
 }
 
 #send {
--- a/addon/content/components/compose/composeWidget.css
+++ b/addon/content/components/compose/composeWidget.css
@@ -72,11 +72,21 @@ input:focus {
 .compose > .buttons > button {
   padding: 0.3em 1em;
   background-color: -moz-dialog;
-  color: -moz-dialogtext;
+  color: #666;
   border: 1px solid ThreeDShadow;
   border-radius: 3px;
 }
 
+@media (prefers-color-scheme: dark) {
+  .compose > .buttons > button {
+    color: -moz-fieldtext;
+  }
+}
+
 .compose > .buttons > button:hover {
   filter: brightness(85%);
+}
+
+.compose > .buttons > button > .icon {
+  margin-inline-end: 0.3em;
 }
--- a/addon/content/components/compose/composeWidget.css
+++ b/addon/content/components/compose/composeWidget.css
@@ -64,12 +64,13 @@ input:focus {
 }
 
 .compose > .buttons {
-  margin-top: 10px;
+  padding-top: 10px;
+  display: flex;
+  justify-content: space-between;
 }
 
 .compose > .buttons > button {
   padding: 0.3em 1em;
-  margin-inline-end: 0.2em;
   background-color: -moz-dialog;
   color: -moz-dialogtext;
   cursor: pointer;
@@ -79,8 +80,4 @@ input:focus {
 
 .compose > .buttons > button:hover {
   filter: brightness(85%);
-}
-
-#send {
-  float: inline-end;
 }
--- a/addon/content/components/compose/composeWidget.jsx
+++ b/addon/content/components/compose/composeWidget.jsx
@@ -7,6 +7,7 @@ import React from "react";
 import * as ReactRedux from "react-redux";
 import { TextArea, TextBox } from "./composeFields.jsx";
 import PropTypes from "prop-types";
+import { SvgIcon } from "../svgIcon.jsx";
 
 export function ComposeWidget({ discard }) {
   const dispatch = ReactRedux.useDispatch();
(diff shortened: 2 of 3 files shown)


Standard8 (changes requested):
Thank you for trying this, sorry for the delay in getting back to you.

As you indicated, the bottom border isn't quite right, and I'm not sure I'd be happy accepting it as-is.

I think what it needs is:

* The `margin-top` changing to `padding-top` [on `.compose > buttons`](https://github.com/thunderbird-conversations/thunderbird-conversations/blob/8b177a0d0fce22c1cd54fa656088aa597e8715d0/addon/content/components/compose/composeWidget.css#L66-L68).
* Removing the [`float: right` on `#send`](https://github.com/thunderbird-conversations/thunderbird-conversations/blob/8b177a0d0fce22c1cd54fa656088aa597e8715d0/addon/content/components/compose/composeWidget.css#L79C1-L81) and replacing with something that still gets the button aligned right.
  * This is needed as otherwise the float is causing the button to overlap padding when it is rendered without float.
  * I think you could do this with the following added to `.compose > .buttons`:

```
display: flex;
justify-content: space-between;
```

I think then it'd give a consistent spacing around the button.

Standard8:
> Future improvements should also include changing links `discard`, `Forward this conversation`, `Print this conversation` to button style. It seems like the UI has been neglected and can be easily improved.

Changing the discard to a button would be reasonable. I don't think I'd agree with changing the forward/print to buttons, I think that would not fit correctly with the general layout/style.

fezjo:
Hello, your suggestions worked flawlessly (as far as I can tell). Thanks!

I also changed the discard link to button and changed the tests accordingly.

Finally, I added pictograms in the form of emojis next to the buttons so it's clear at a glance what the actions does. This is again just a suggestion, and may not be the best solution. Firstly, the trash can may be misleading as the email is not moved to _Trash_ but discarded. Also, I am not sure if all clients/fonts support these emojis and Thunderbird uses SVG icons. I would also prefer to use the same icons as Thunderbird, but this seemed like a quick and trivial initial improvement. Again, if you don't like the style or have suggestions, feel free to say or change anything.

image

Standard8:
For the icons, if you rebase on the latest, you'll see that I've just added a `send` icon to `material-icons.svg`. There's already a `delete_forever` icon, that I think we could use for the discard.

In theory, you should be able to use the [`SvgIcon` class to display the icons](https://github.com/thunderbird-conversations/thunderbird-conversations/blob/8b177a0d0fce22c1cd54fa656088aa597e8715d0/addon/content/components/message/messageActionButton.jsx#L76).

fezjo:
Here's the version with SVG icons. I put the send icon to the right because it looked weird on the left and now it forms a big arrow with the text.

Also, I put a `nbsp` in between. As you can maybe tell, I don't really know what I am doing with web stuff :D

image

addon/content/components/compose/composeWidget.css
```diff
@@ -64,18 +64,19 @@ input:focus {
...
+  justify-content: space-between;
 }
 
 .compose > .buttons > button {
-  padding: 0 5px;
-  margin-inline-end: 3px;
-  background-color: white;
+  padding: 0.3em 1em;
+  background-color: -moz-dialog;
+  color: -moz-dialogtext;
```
Standard8, line 75: This seems fine in dark mode (though maybe a little bright?), but in light mode it is too harsh compared to the rest of the UI.

Normally in light mode we're using `#666`: https://github.com/thunderbird-conversations/thunderbird-conversations/blob/1f5a213c734bafc535587a6d65cf54166865a4f4/addon/content/conversation.css#L152

So I think we want to have `#666` as the default and `-moz-dialogtext` as the `(prefers-color-scheme: dark)` option.

addon/content/components/compose/composeWidget.jsx
```diff
@@ -99,11 +100,11 @@ export function ComposeWidget({ discard }) {
...
       <div className="buttons">
-        {discard && (
-          <a className="link" onClick={discard}>
-            {browser.i18n.getMessage("compose.discard")}
-          </a>
-        )}
+        <button id="discard" onClick={discard} disabled={!discard}>
+          <SvgIcon ariaHidden={true} hash="delete_forever" />
+          &nbsp;
+          {browser.i18n.getMessage("compose.discard")}
```
Standard8, line 106: For consistency with the rest of the UI, let's make both of these buttons have the icon first, and then the text.

If you drop the `&nbsp;` you can use `margin-inline-end: 0.1em;` within a `.compose > .buttons > button > .icon` to apply a bit of spacing.

Standard8 (changes requested):
Thank you for the updates, I think we're almost there, just a few more tweaks.

Standard8:
@fezjo did you see my comments above, would you be able to update for those? If not, would you mind if I addressed those?

I'm planning another update soon, so it would be good to get this landed.

addon/content/components/compose/composeWidget.css
fezjo: Well if that's what you normally do, then of course use 666.

However, maybe all of your UI should eventually be consistent with the user theme. But that's work for other PR...

addon/content/components/compose/composeWidget.jsx
fezjo: Ok

fezjo:
Hello, I saw your comments but had to focus on other stuff. Sorry for not replying sooner. I can find time for this during the weekend, but I don't mind you addressing them. Thank you for your guidance, I learned a bit about UI :)

addon/content/components/compose/composeWidget.css
Standard8: Yeah, let's be consistent for now, and we can explore different options in later options.

Standard8 (approved):
Thank you for all the updates, this looks good. Nice to get this fixed.

PR #2160, opened 2024-11-01
Update messages.json [Ukrainian translation]

artpoli:


Standard8 (commented):
Thank you.

Just so that you know - I got the translation for `attachments.delete.warning` directly from Thunderbird, so you might want to propose an update to Thunderbird as well.

PR #2165, opened 2024-11-18
Fix issue 1611: Add native theme support (& dark mode) to the message display

philia897:
This commit mitigates the issue of dark mode adaptability.

Adding one feature to the options to allow user to enable dark mode. Then the content in the conversation nodes will be adapted.

The solutions is inspired by this comment:
https://github.com/thunderbird-conversations/thunderbird-conversations/issues/1611#issuecomment-2470675052

PS. 
1. I modified all the messages.json under locale, but all are English, feel free to change to the correct language.
2. When I tried to build the project, build.sh does not work well because of the error of `find . -E ...`. The regex expression seems not work, fixed it as well.

### DIFF
--- a/addon/content/components/message/messageIFrame.mjs
+++ b/addon/content/components/message/messageIFrame.mjs
@@ -571,9 +571,8 @@ export class MessageIFrame extends React.Component {
   }
 
   injectCss(iframeDoc) {
-    // !important because messageContents.css is appended after us when the html
-    // is rendered
-    return [
+    // Base CSS rules
+    let cssRules = [
       'blockquote[type="cite"] {',
       "  border-right-width: 0px;",
       "  border-left: 1px #ccc solid;",
@@ -583,6 +582,21 @@ export class MessageIFrame extends React.Component {
       "  height: auto;",
       "}",
     ];
+
+    // Additional CSS for dark mode
+    cssRules.push(
+      `@media not print and (prefers-color-scheme: dark){
+          body {
+            filter: invert(100%) hue-rotate(180deg) !important;
+            background: rgb(28, 27, 34) !important;
+          }
+          body :is(img, [style*="background-image:"]:not([style*="background-image: none"]), [background*="."], g-emoji) {
+            filter: invert(100%) hue-rotate(180deg) !important;
+          }
+      }`
+    );
+
+    return cssRules;
   }
 
   async _onDOMLoaded(event) {
--- a/addon/content/components/message/messageIFrame.mjs
+++ b/addon/content/components/message/messageIFrame.mjs
@@ -585,7 +585,7 @@ export class MessageIFrame extends React.Component {
 
     // Additional CSS for dark mode
     cssRules.push(
-      `@media not print and (prefers-color-scheme: dark){
+      `@media screen and (prefers-color-scheme: dark) {
           body {
             filter: invert(100%) hue-rotate(180deg) !important;
             background: rgb(28, 27, 34) !important;


philia897:
Seems the tests did not run automatically? I did `npm test` locally and it passed.

Pacman99:
Just tried this out and it works pretty well. I'm attaching some screenshots.
[Text Message](https://github.com/user-attachments/assets/dcc2beba-270d-425c-90f0-70e730532eab)
[HTML Message](https://github.com/user-attachments/assets/6ca6c7bf-9c7a-4312-9e7c-bb180d791faf)

janhenke:
Hi,
I am exactly looking for this one (getting dark mode to work tother with this addon). How can we get this one merged and released? Who is able to review and merge this PR?

Standard8:
> How can we get this one merged and released? Who is able to review and merge this PR?

I need to review it. This is to acknowledge that I've seen it. Unfortunately, I'm going to be busy until the end of the year and won't have time to dig through until then, so it'll be at least a few more weeks yet.

lmkra:
FYI: it seems that modified build.sh does not work correctly on Mac OS.
After reverting this script to the original one, everything builds correctly.

philia897:
> FYI: it seems that modified build.sh does not work correctly on Mac OS. After reverting this script to the original one, everything builds correctly.

That shall be a conflict between Linux and Mac I guess. I developed upon Arch Linux, and the original regex expression does not match anything.

But anyway, that is my fault. I think we shall keep the original version of build.sh for this pull request anyway, to avoid unexpected conflicts.

Standard8 (changes requested):
Thank you for working this. I'm sorry it has taken a while for me to be able to look at it.

I think the general idea for applying dark mode is fine, but we should simplify how it is applied. Rather than adding a preference, I think we should be able to automatically detect what's happening.

For example, in messageIFrame you can replace the `if (isDarkMode) {` checks with `if (window.matchMedia("(prefers-color-scheme: dark)").matches) {`. It won't dynamically change (although we could add a listener), but it would at least update when you select a different message.

If we want to supply a button to allow the user to force toggle the display of a single message, then I think we could probably do that. The icon set we use has a contrast icon, so we could probably use that.

Please also drop the other changes from this PR. Allowing the sorting of Conversations would need a lot more work - especially for the dynamic insertion of messages and how to handle quick reply. Generally it is better to keep individual PRs fixing a single issue to make it easier to understand the changes and to review.

philia897:
> Thank you for working this. I'm sorry it has taken a while for me to be able to look at it.
> 
> I think the general idea for applying dark mode is fine, but we should simplify how it is applied. Rather than adding a preference, I think we should be able to automatically detect what's happening.
> 
> For example, in messageIFrame you can replace the `if (isDarkMode) {` checks with `if (window.matchMedia("(prefers-color-scheme: dark)").matches) {`. It won't dynamically change (although we could add a listener), but it would at least update when you select a different message.
> 
> If we want to supply a button to allow the user to force toggle the display of a single message, then I think we could probably do that. The icon set we use has a contrast icon, so we could probably use that.
> 
> Please also drop the other changes from this PR. Allowing the sorting of Conversations would need a lot more work - especially for the dynamic insertion of messages and how to handle quick reply. Generally it is better to keep individual PRs fixing a single issue to make it easier to understand the changes and to review.

Thanks for your suggestions and review @Standard8 . I will add this to my schedule. Though these days I'm a little busy so can not guarantee the progress. I will find time to do this.

Hopefully I will commit something new in two weeks.

SylwesterZarebski:
Is there any chance to accept this change and release new addon version soon?

philia897:
> Is there any chance to accept this PR and release new addon version soon?

Well, I'm still waiting for @Standard8 to review the changes. Hopefully he will update soon but I'm not sure when the new version will be released.

jojobg:
@philia897 With the compatibility for 137 fixed this PR will make the add-on "whole" again in the new fancy dark mode. Thank you in advance.

ryanwwest:
It's not clear to me from the thread if the new native Dark mode within messages (and the toggle on/off in the message header) released in Thunderbird 136 will be used here. Would it be good to adopt that or integrate with that? https://blog.thunderbird.net/2025/03/thunderbird-release-channel-update/

janhenke:
> It's not clear to me from the thread if the new native Dark mode within messages (and the toggle on/off in the message header) released in Thunderbird 136 will be used here. Would it be good to adopt that or integrate with that? https://blog.thunderbird.net/2025/03/thunderbird-release-channel-update/

The toggle is unavailable if you use this plugin, since it replaces the whole message frame. Therefore, this feature is highly needed and must be implemented in the plugin.

Standard8 (approved):
Thank you for updating this. I've just tweaked it a bit so that the media is correctly applied - previously it was applying the dark mode to the light mode as well.

There's still some extra things we need to do though, I've got a few locally and I'll comment about the rest in the issue.

Issue #1900, opened 2022-09-20
Sort messages in conversation with the newest at the top


reaseno:
How can I sort messages in the messages panel so that the newest messages are at the top. Exactly in the reverse order of the current standard. I have not been able to find a setting for this unfortunately.

If it is not possible, I would like to make a feature request for it.

Standard8:
[As per the FAQ](https://github.com/thunderbird-conversations/thunderbird-conversations/wiki/FAQ#message-display), this generally hasn't been done because of the additional complexity that it would create in the code.

I think it might be a little more viable now with refactors of the code base but I would still be concerned that it would be difficult to support well (e.g. especially when inserting messages into the currently displayed conversation - could cause unwanted scrolling etc).

For me, this is a low priority compared to all the other issues and feature requests that come in.

godzfire:
I'm shocked that there hasn't been more responses to this, but not being able to sort and see most recent replies at the top like Outlook is an absolute deal breaker for me. It seems counterintuitive you have to scroll down large email conversations just to see the most recent email.

Vigilans:
> I'm shocked that there hasn't been more responses to this, but not being able to sort and see most recent replies at the top like Outlook is an absolute deal breaker for me. It seems counterintuitive you have to scroll down large email conversations just to see the most recent email.

Yes, this just prevents me from adopting this extension in the first place despite all other features and merits.

godzfire:
I am hoping this can be re-evaluated to inclusion. It's stated "For me, this is a low priority compared to all the other issues and feature requests that come in.", however my previous reply in this thread got 11 thumbs up by fellow members feeling the same way.

This feature is something that is highly wanted.

Standard8:
> This feature is something that is highly wanted.

I do understand that some people want this feature. A lot more people have signalled they want dark mode, some people want encrypted email working again (a major regression of the last Thunderbird upgrade).

As I stated before, the UX for this feature is not necessarily trivial to implement and needs thinking about and careful testing.

Unfortunately I barely get enough time at the moment to keep the add-on going, let alone think about new features. When I said this was low priority, I'm just being honest with you that I simply don't have time for it.

I'd be happy for someone to take this on, but of course, that would need to be someone that can write the appropriate code.

philia897:
@Standard8 

I've made a simple fix to this issue (See in this pull request #2165 ). It simply reverse the order of messages without changing other things. This feature can be enabled in options.

[...]

PR #2366, opened 2026-04-17
[codex] Add optional reverse conversation order

Symplify62:
## Summary
- add an optional setting to display the newest conversation messages first
- keep the underlying conversation data in chronological order while deriving display order in the React layer
- keep Quick Reply attached to the newest logical message when reverse order is enabled

## Why
Issue #1900 requests an option to show the newest message at the top of a conversation. The existing implementation assumes chronological storage order in several places, so this change keeps the data model intact and limits the feature to display-order behavior plus the related Quick Reply placement.

## Testing
- NODE_OPTIONS='--import "./addon/tests/setup.mjs"' node --test addon/tests/prefs.test.mjs addon/tests/messageList.test.mjs addon/tests/quickReply.test.mjs addon/tests/options.test.mjs
- npx eslint addon/background/prefs.mjs addon/options/options.mjs addon/_locales/en/messages.json addon/content/reducer/reducerSummary.mjs addon/content/reducer/controllerActions.mjs addon/content/components/message/messageList.mjs addon/tests/prefs.test.mjs addon/tests/messageList.test.mjs
- npm run build

### DIFF
--- a/addon/content/components/message/messageList.mjs
+++ b/addon/content/components/message/messageList.mjs
@@ -6,6 +6,17 @@ import React from "react";
 import * as ReactRedux from "react-redux";
 import { Message } from "./message.mjs";
 
+function getDisplayMessages(msgData, reverseConversationOrder) {
+  if (!msgData) {
+    return [];
+  }
+  return reverseConversationOrder ? [...msgData].reverse() : msgData;
+}
+
+function getQuickReplyTargetId(msgData) {
+  return msgData?.[msgData.length - 1]?.id;
+}
+
 /**
  * Handles display of the list of messages.
  *
@@ -19,6 +30,11 @@ function _MessageList({ dispatch, messages, summary }) {
   // can be called on them in response to a `advanceMessage()`
   // call. The actual ref is stored in `React.useRef().current`
   const { current: childRefs } = React.useRef([]);
+  const displayMessages = getDisplayMessages(
+    messages.msgData,
+    summary.prefs.reverseConversationOrder
+  );
+  const quickReplyTargetId = getQuickReplyTargetId(messages.msgData);
 
   function setRef(index, ref) {
     childRefs[index] = ref;
@@ -35,8 +51,7 @@ function _MessageList({ dispatch, messages, summary }) {
   return React.createElement(
     "ul",
     { id: "messageList" },
-    !!messages.msgData &&
-      messages.msgData.map((message, index) =>
+    displayMessages.map((message, index) =>
         React.createElement(Message, {
           key: index,
           autoMarkAsRead: summary.autoMarkAsRead,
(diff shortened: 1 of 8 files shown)


addon/content/components/message/messageList.mjs
```diff
@@ -19,6 +30,11 @@ function _MessageList({ dispatch, messages, summary }) {
   // can be called on them in response to a `advanceMessage()`
   // call. The actual ref is stored in `React.useRef().current`
   const { current: childRefs } = React.useRef([]);
+  const displayMessages = getDisplayMessages(
+    messages.msgData,
+    summary.prefs.reverseConversationOrder
+  );
+  const quickReplyTargetId = getQuickReplyTargetId(messages.msgData);
```
Standard8, line 37: There's no need for the intermediate `getQuickReplyTargetId` function, it is only called here and doesn't give any additional information.

Standard8, line 37: These should be be defined just before they're used - i.e. just before the return statement.

addon/content/components/message/messageList.mjs
```diff
@@ -6,6 +6,17 @@ import React from "react";
 import * as ReactRedux from "react-redux";
 import { Message } from "./message.mjs";
 
+function getDisplayMessages(msgData, reverseConversationOrder) {
+  if (!msgData) {
+    return [];
+  }
+  return reverseConversationOrder ? [...msgData].reverse() : msgData;
```
Standard8, line 13: Using `toReversed()` should be more optimal, and avoids the separate copy and then reverse.

addon/content/components/message/messageList.mjs
```diff
@@ -6,6 +6,17 @@ import React from "react";
...
+}
+
+function getQuickReplyTargetId(msgData) {
+  return msgData?.[msgData.length - 1]?.id;
```
Standard8, line 17: Why do we need the additional check for if `msgData?.[msgData.length - 1]` exists or not? The existing code doesn't have it, it seems unnecessary.

addon/tests/messageList.test.mjs
```diff
@@ -0,0 +1,109 @@
...
+import React from "react";
+import * as RTK from "@reduxjs/toolkit";
+import * as ReactRedux from "react-redux";
+import { MessageList } from "../content/components/message/messageList.mjs";
+import { Message } from "../content/components/message/message.mjs";
+import { conversationApp } from "../content/reducer/reducer.mjs";
+import { messageActions } from "../content/reducer/reducerMessages.mjs";
+import { summaryActions } from "../content/reducer/reducerSummary.mjs";
+
+function stubMessageRendering(t) {
```
Standard8, line 18: This seems an unnecessary amount of stubbing for what we usually do, and looks to kinda make the tests almost pointless.

See message.test.mjs for example where we use helpers to create fake messages for display.

Standard8 (changes requested):
Thank you for proposing this, I'm sorry it has taken a while for me to get to look at it. The basic idea looks reasonable, but there's various optimisations and improvements that could be applied. Some of these feel like unnecessary mistakes that a human wouldn't have made. If you have submitted without reading and understanding it yourself, please consider reviewing patches yourself in future to avoid unnecessary work for those reviewing your patches.

Standard8:
Before updating, please also see the new AGENTS.md and CONTRIBUTING.md on the latest main.

PR #2391, opened 2026-07-11
fix: enable attachment drag and drop support

MatrixNeoKozak:
Fetches the File object for all non-deleted attachments, sets up standard HTML5 and gecko-specific drag and drop properties on dragstart, and marks the attachment list item as draggable.

### DIFF
--- a/addon/content/components/message/attachments.mjs
+++ b/addon/content/components/message/attachments.mjs
@@ -109,32 +109,56 @@ function Attachment({
     );
   }
 
-  // TODO: Fix drag n drop of attachments.
-  // onDragStart(event) {
-  //   let info;
-  //   if (/(^file:|&filename=)/.test(this.props.url)) {
-  //     info = this.props.url;
-  //   } else {
-  //     info =
-  //       this.props.url +
-  //       "&type=" +
-  //       this.props.contentType +
-  //       "&filename=" +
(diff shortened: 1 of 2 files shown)


Standard8 (changes requested):
This change doesn't work:

```
Uncaught ReferenceError: can't access lexical declaration 'isImage' before initialization
    Attachment chrome://conversations/content/content/stub.bundle.js:422
```

This happens as soon as I try and load a message with an attachment. Even if I fix that, then messages still don't load.

I'm glad you are trying, but please ensure that you manually test all changes before submission (especially if you are using AI) - especially when not including tests. Please do not waste reviewer's time.

I will say I'm also not happy about loading every attachment for messages as soon as you open the message - some attachments can be large, and loading them into memory isn't acceptable, especially when drag and drop might not happen straight away. It might also force them to be downloaded when users don't want that. We do it for images and get away with it, though I'm not all that pleased with how they work.

This is why this problem is hard to fix, [as I alluded to here](https://github.com/thunderbird-conversations/thunderbird-conversations/issues/1722#issuecomment-1072237473). I'm happy for other ways to be explored, but I don't think we want to load into memory for non-images.

Standard8:
Before updating, please also see the new AGENTS.md and CONTRIBUTING.md on the latest main.

PR #2396, opened 2026-07-20
Wrap conversation subjects in vertical and separate views

### LINKED ISSUE #2049: Feature suggestion:
Allow full subject to be shown at the top of the conversation. (reducing font size to hopefully get more of it displayed is not an accessible solution)

Currently, subjects don't need to be too long before you can't read them without a slow pointer-hover-wait-for-tooltip action

Currently

screenshot of a conversation with a long-ish subject that is truncated with elipsis. An arrow points to this.

Suggestion: allow as much space as needed to fit the subject.

same screenshot mocked up to show the grey box that contains the subject is now expanded vertically to fit the whole subject in which wraps over 2 lines. The rest is identical.

An email's subject is a really key part of the message, and should not be cropped off, in my opinion. And allowing a tiny bit more vertical space for this, once, at the top of the conversation list seems like a small price to pay for being able to read the subject.

Thanks for the excellent extension.

AysajanE:
## Summary

- preserve single-line subject truncation in the classic three-pane layout
- show the full subject across lines in vertical layout, separate tabs, and standalone windows
- refresh the cached conversation layout after Thunderbird layout changes
- cover both view-specific wrapping and live classic/vertical transitions

## Checks

- `OPENSSL_CONF=/dev/null node --test addon/tests/conversationHeader.test.mjs addon/tests/reducerController.test.mjs` (base failed and corrected commit passed in the clean network-off verifier)
- `npm test` (90/90 tests passed, including lint, TypeScript, formatting, and Node tests)
- `npm run build` (webpack and XPI build completed; existing raw-component copy warnings remain)
- Thunderbird 152.0.1 manual UI matrix: classic, vertical, separate-tab, and standalone-window views passed; live classic -> vertical -> classic updates passed in the same open conversation

Fixes #2049

---
AI assistance was used. This change was reviewed by Northset, and I accept responsibility for this submission.

### Verification

[Northset proof-of-pass receipt M-1025](https://northset-oss.github.io/verification-pilot/receipts/M-1025/)  
Contributor self-run; not maintainer verification.

### DIFF
--- a/addon/content/components/conversation/conversationHeader.mjs
+++ b/addon/content/components/conversation/conversationHeader.mjs
@@ -409,6 +409,12 @@ export class ConversationHeader extends HTMLElement {
     if (this.#linkifiedSubject.getAttribute("subject") != summary.subject) {
       this.#linkifiedSubject.setAttribute("subject", summary.subject);
     }
+    this.#linkifiedSubject.toggleAttribute(
+      "wrap",
+      summary.isVerticalLayout === true ||
+        summary.isInTab === true ||
+        summary.isStandalone === true
+    );
     if (
       this.#convActionButtons.getAttribute("darkreaderenabled") !=
       summary.darkReaderEnabled.toString()
--- a/addon/content/conversation.css
+++ b/addon/content/conversation.css
@@ -146,6 +146,14 @@ a img {
   text-overflow: ellipsis;
 }
 
+:host([wrap]) .subject {
+  height: auto;
+  overflow: visible;
+  overflow-wrap: anywhere;
+  white-space: normal;
+  text-overflow: clip;
+}
+
 linkified-subject {
   flex: 1;
   overflow: hidden;
--- a/addon/content/reducer/controllerActions.mjs
+++ b/addon/content/reducer/controllerActions.mjs
@@ -75,10 +75,17 @@ export const controllerActions = {
         tabId = isStandalone ? null : BrowserSim.getTabId(topWin, window);
       }
 
+      let isVerticalLayout = false;
+      if (!isInTab && tabId != null) {
+        isVerticalLayout =
+          (await browser.mailTabs.get(tabId)).layout === "vertical";
+      }
+
       await dispatch(
         summaryActions.setConversationState({
           isInTab,
           isStandalone,
+          isVerticalLayout,
           tabId,
           windowId,
         })
--- a/addon/content/reducer/controllerActions.mjs
+++ b/addon/content/reducer/controllerActions.mjs
@@ -160,6 +160,22 @@ export const controllerActions = {
     };
   },
 
+  updateConversationLayout() {
+    return async (dispatch, getState) => {
+      const { isInTab, isStandalone, tabId } = getState().summary;
+      if (isInTab || isStandalone || tabId == null) {
+        return;
+      }
+
+      const { layout } = await browser.mailTabs.get(tabId);
+      dispatch(
+        summarySlice.actions.setVerticalLayout({
+          isVerticalLayout: layout === "vertical",
+        })
+      );
+    };
+  },
+
   /**
    * Handles potentially marking a conversation as read.
    */
@@ -433,6 +449,10 @@ function setupListeners(dispatch, getState) {
   let port = browser.runtime.connect({ name: "externalMessages" });
   let externalMessagesListener = onExternalMessages.bind(this, dispatch);
   port.onMessage.addListener(externalMessagesListener);
+  const layoutChangedListener = () => {
+    dispatch(controllerActions.updateConversationLayout()).catch(console.error);
+  };
+  window.addEventListener("resize", layoutChangedListener);
 
   window.addEventListener(
     "unload",
@@ -459,6 +479,7 @@ function setupListeners(dispatch, getState) {
       browser.convOpenPgp.onSMIMEReload.removeListener(smimeReloadListener);
       port.onMessage.removeListener(externalMessagesListener);
       port.disconnect();
+      window.removeEventListener("resize", layoutChangedListener);
     },
     { once: true }
   );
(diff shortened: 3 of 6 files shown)


Standard8 (changes requested):
Presumably the idea for this came from #2049 ? So that should be linked.

> No manual Thunderbird UI verification was performed.

If you're not going to bother to review your changes manually AND test them, then I'm not going to bother to review your path.

AysajanE:
Thanks for calling this out—you were right that the original PR was missing both the issue link and manual verification.

This is for #2049, and the PR now includes `Fixes #2049`.

I rebuilt commit `c9d76b2` as an XPI and installed it in a clean Thunderbird 152.0.1 test profile with a local synthetic two-message conversation. I manually verified:

- classic three-pane: the subject remains single-line and truncates;
- vertical layout: the full subject wraps;
- separate tab: the full subject wraps;
- live vertical → classic → vertical switching in the same open conversation refreshes correctly.

That manual pass also confirmed why `c9d76b2` is needed: a stale pre-correction XPI reproduced the cached-layout failure, while the XPI built from `c9d76b2` passed the same transition.

I've attached the matched classic, vertical, and separate-tab captures below. Would you mind taking another look?

thunderbird-152-classic-layout

thunderbird-152-vertical-layout

thunderbird-152-separate-tab

Standard8:
I asked for manual testing, i.e. the human must test this. Please read the AGENTS.md on latest main before responding further.

AysajanE:
@Standard8 I am working on it (Aysajan Eziz, a human)

AysajanE:
HI @Standard8. I personally drove the complete manual test, including the uninterrupted Classic → Vertical → Classic → Vertical sequence, and captured each screenshot one by one. The following are the artifacts, along with the screenshots captured. 

  - Date: July 25, 2026
  - Commit: c9d76b2df2b280636a55422d522fde38f31edcc1
  - Thunderbird: 152.0.1
  - Conversations: 4.3.10
  - Classic: truncated to one line with ellipsis
  - Vertical: full subject wrapped
  - Continuous Classic → Vertical → Classic → Vertical: passed without reload or reselection
  - Separate tab: wrapped
  - Standalone window: wrapped
  
  
07-standalone-window
06-separate-tab
05-vertical-after-second-live-switch
04-classic-after-live-switch-back
03-vertical-after-live-switch
02-classic-before-switch
01-addon-version
00-runtime-temporary-addon
01a-addon-version
01-temporary-addon
00-thunderbird-version

addon/content/components/conversation/conversationHeader.mjs
```diff
@@ -409,6 +409,12 @@ export class ConversationHeader extends HTMLElement {
     if (this.#linkifiedSubject.getAttribute("subject") != summary.subject) {
       this.#linkifiedSubject.setAttribute("subject", summary.subject);
     }
+    this.#linkifiedSubject.toggleAttribute(
+      "wrap",
+      summary.isVerticalLayout === true ||
+        summary.isInTab === true ||
+        summary.isStandalone === true
```
Standard8, line 416: Please use non-strict equals. There's generally not much to be gained from being strict unless you're calling out there's a specific state.

addon/content/reducer/controllerActions.mjs
```diff
@@ -75,10 +75,17 @@ export const controllerActions = {
         tabId = isStandalone ? null : BrowserSim.getTabId(topWin, window);
       }
 
+      let isVerticalLayout = false;
+      if (!isInTab && tabId != null) {
+        isVerticalLayout =
+          (await browser.mailTabs.get(tabId)).layout === "vertical";
```
Standard8, line 81: non-strict equals (also please fix the other places in the patch)

addon/content/conversation.css
```diff
@@ -146,6 +146,14 @@ a img {
   text-overflow: ellipsis;
 }
 
+:host([wrap]) .subject {
+  height: auto;
```
Standard8: I think we should limit the height, in case of extremely long titles, which can mess up the whole display.

Maybe adding `max-height: 5lh;` would be about right?

addon/content/conversation.css
```diff
@@ -146,6 +146,14 @@ a img {
...
+  overflow: visible;
+  overflow-wrap: anywhere;
+  white-space: normal;
+  text-overflow: clip;
+}
+
```
Standard8, line 156: I think we should still limit the overall height, in case you do get an extremely long subject, which messes up the display so that you can't read the body at all.

I think maybe 5 lines worth is enough?

Looking around, we'd need something like `line-clamp` but support for that isn't yet enabled, however it does appear that we can use `-webkit-line-clamp`.

So I suggest that we go with:

```css
:host([wrap]) .subject {
  height: auto;
  -webkit-line-clamp: 5;
  /* line-clamp is not yet supported in Thunderbird (bug 1540681) */
  line-clamp: 5;
  white-space: normal;
}
```

This should also be future proof for when `line-clamp` gets turned on.

addon/tests/conversationHeader.test.mjs
```diff
@@ -0,0 +1,68 @@
...
+
+  function setView(header, view = {}) {
+    header.setData(
+      {
+        darkReaderEnabled: false,
+        isInTab: false,
+        isStandalone: false,
+        isVerticalLayout: false,
+        loading: false,
+        subject: "A subject long enough to need more than one line",
```
Standard8, line 33: This is misleading because we're not testing the actual wrapping of the subject. That's fine, just rename this to `A subject` or something like that.

Standard8 (changes requested):
@AysajanE Thank you very much for the updates and the testing. I'm sorry I haven't got around to looking at this earlier, summer has been busy.

There's a few comments in-line, I think the major one is limiting the height to some extent (which I've given a solution for), the other ones are minor.

